PR #1, opened 2020-09-17
Move files to typescript

acoates-ms:
Convert the JavaScript files into typescript.

- Basically just renaming files from js -> ts or tsx depending on if they require the x.
- Also removed the flowconfig file, since we're not going to be using flow in this project.
- Also bumped some of the lock file (I just nuked the whole file and regenerated it, but the important update I wanted was the newer CLI which has a fix to use the users default browser instead of trying to push chrome on people.
- My local config ran prettier automatically on some files which caused some formatting difference.  Looks like there is a prettier config in the repo, so I suspect there should be some future work to help people get setup in a way that formats correctly automattically.


kmelmon (approved):
:shipit:

PR #2, opened 2020-09-17
Get lint clean

acoates-ms:
Fixes the existing lint errors
Moves WelcomePage and CheckBoxExamplePage to FunctionalComponents
Makes vscode suggest the eslint and prettier addons
Configures vscode to run format and lint on save

- I think this gets the repo to a reasonable JS dev experience


kmelmon (approved):
:shipit:

PR #7, opened 2020-09-18
Update to 0.63

acoates-ms:
This updates the version of react-native and react-native-windows to 0.63.


kmelmon:
>     PackageProviders().Append(winrt::CheckboxWindows::ReactPackageProvider());

Auto-linking should be taking care of this now, so shouldn't we remove this?
I also added #include <winrt/CheckboxWindows.h> to pch.h, which we should also be able to remove.

---
Refers to: windows/rngallery/App.cpp:43 in c07c6a4. [](commit_id = c07c6a4769ad21e91cd2d15329a7c05e33e2f0b6, deletion_comment = False)

acoates-ms:
Just tried it, and doing that gets me a redbox when visiting the CheckBox page, due to a missing view manager.  
Not sure if autolinking works with the checkbox module.. or whats up with that.

acoates-ms:
My instinct would be to merge this as is, then maybe have Jon or someone look at why autolinking isn't working.

kmelmon:
k let's do that

---
In reply to: [695091709](https://github.com/microsoft/RNGallery/pull/7#issuecomment-695091709) [](ancestors = 695091709)

kmelmon (approved):
:shipit:

PR #15, opened 2020-11-23
Add code formatting and example components

acoates-ms:
Added a `Code` component, which parses typescript code and renders native RN text elements to render formatted code.
Added a `Page` and `Example` component to share between example pages.
Added a bunch more examples to the checkbox example page.

Here is what the formatted code component looks like:
rng


chiaramooney:
Few thoughts:
Eventually when we add copying the code snippet to one's clipboard back in, it might not make sense for the import line and jsx for the component to be in one text block and copied all together, since these pieces will be placed in separate areas of the .js/.tsx files. Maybe in future the import line and the jsx for the component could be separate string inputs for Example and rendered in separate stacked blocks, so eventually they can be each copied independently. 

Looks like in the XAML Controls Gallery the margins are even across the left and right of the page. I like the look of that layout slightly better, but that's just my personal preference. 

Would be cool in future to add `Toggle Theme` button like in the XAML Controls Gallery and regular/light mode syntax highlighting to Code so users go also view in regular/light mode too.

chiaramooney (approved):

PR #19, opened 2020-12-14
Upgrade to latest version of react-native-windows

tritao:
This fixes a build issue with Folly and latest MSVC version due to missing compiler builtins.

Error message is as follows:

`folly\portability\Builtins.h(109,15): error C3861: '__builtin_ctzll': identifier not found`

Fixed upstream in https://github.com/facebook/folly/commit/7dcf1524c340e410fe95170b4a920de54d666214.

Uses the ^ prefix for the package specifier so that yarn picks minor and patch version updates automatically on install.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,12 +20,12 @@
     "@react-navigation/stack": "^5.9.0",
     "lowlight": "^1.17.0",
     "react": "16.13.1",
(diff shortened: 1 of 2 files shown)


chiaramooney:
Great thanks for spotting this. Change added in another PR that was just merged.

PR #28, opened 2020-12-21
Add Theme Switch

chiaramooney:
This PR is code that was originally in the Add Settings Page PR that has now been separated out for a clearer flow. 

Added Light Theme support for Code Snippets, tried to match coloring of syntax highlighting in XAML Controls Gallery. Added global state for theme. Using dropdown in Settings, users can select which theme they want shown for code snippets. 

Tried to add Theme toggle for entire app but DefaultTheme and DarkTheme from @react-navigation/native seem to only change background color and will NOT change text coloring or coloring of the app window. Only way to view dark or light theme for the app correctly is when the user's Windows system has selected for apps to display in dark or light theme. 

**_Example of Light Theme Code Snippet_**
image

**_Example of Dark Theme Code Snippet_**
image

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,5 +5,10 @@
 import {AppRegistry} from 'react-native';
 import App from './src/App';
 import {name as appName} from './app.json';
+import {setGlobal} from 'reactn';
+
+setGlobal({
+  theme: 'system',
+});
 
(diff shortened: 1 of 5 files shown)


kmelmon (approved):
:shipit:

PR #32, opened 2020-12-22
Add Build to PR, Fix Yarn Issue

chiaramooney:
`yarn install` was being skipped with previous main.yml version for some commits. Change now always runs yarn install. 

Added native build to pr.yml so for PR checks to pass native build will run and must succeed.


kmelmon (approved):
:shipit:

PR #37, opened 2021-01-08
Add Text Component Page

chiaramooney:
Added page for Text core component. 

**_Snippet of Text Page_**
image

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,6 +1,6 @@
 name: PR
 
-on: [pull_request_target]
+on: [pull_request]
 
 jobs:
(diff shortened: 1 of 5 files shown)


kmelmon:
This looks great!  

Just one request:  I'd also like to see an example of nested Text, illustrating how you can set different styles (eg first do italicized, then do bold in the nested Text, or something along those lines).

kmelmon (approved):
:shipit:

chiaramooney:
> This looks great!
> 
> Just one request: I'd also like to see an example of nested Text, illustrating how you can set different styles (eg first do italicized, then do bold in the nested Text, or something along those lines).

@kmelmon Nested Text example now added!

PR #43, opened 2021-01-14
Resolve Warnings with Lists

chiaramooney:
Warning message was being fired throughout the app talking about the need for components rendered in a list to have a unique "key". Updated LinkTile and Code files so that all components rendered with a call to the map function have a defined "key" value. This resolved the warning messages. 

**_Screenshot of Warning Message_**
image


src/components/Code.tsx
```diff
@@ -262,9 +262,9 @@ function renderLowLightNode(node: lowlight.AST.Unist.Node) {
 
 function renderLowLightTree(tree: lowlight.HastNode[]) {
   if (tree.length === 1) {
-    return renderLowLightNode(tree[0]);
+    return renderLowLightNode(tree[0], 0);
   } else {
-    return <>{tree.map(renderLowLightNode)}</>;
+    return <>{tree.map((node, index) => renderLowLightNode(node, index))}</>;
```
kmelmon, line 267: >index [](start = 30, length = 5)

It looks like this is using the index of an item in a list as the key.  That isn't reliable the key must stay unique for the lifetime of the list item.  If you use the index, the diff'ing engine can be tricked into not producing a new item which would be a bug.  It can think item n is still the same even in the case that item n is a new item (either through an insert or a remove).  The guidance I've seen in the past is to use the text from the item as that is usually unique.  Note that I'm not an expert on generating unique keys, perhaps there's an internet article on best practices.  Here's an article that might help:  [https://stackoverflow.com/questions/39549424/how-to-create-unique-keys-for-react-elements](https://stackoverflow.com/questions/39549424/how-to-create-unique-keys-for-react-elements)

kmelmon:
Note that I'll be OOF on vacation monday - wednesday.  If you're blocked on checkin, feel free to loop in other RNW devs.

src/components/Code.tsx
chrisglein: https://reactjs.org/docs/lists-and-keys.html#keys

chrisglein: There's some more info linked from the [React.js docs on keys](https://reactjs.org/docs/lists-and-keys.html#keys).

chiaramooney: Resolved.

src/components/Code.tsx
```diff
@@ -195,7 +195,11 @@ const lightStyles = StyleSheet.create({
   'hljs-attr': {color: lightColors.text}, // Attributes within JSX Tags
 });
 
-function renderLowLightNode(node: lowlight.AST.Unist.Node) {
+function createKey(value: string) {
+  return `${value}_${new Date().getTime()}`;
```
chrisglein, line 199: Won't this ensure that the keys are always different? (as each render will append the index to a new date)
That would result in always going down the slow path of throwing away the old UI and creating new UI.

chiaramooney: Ah I see. I removed the timestamp from the key and replaced it with the text value or element tagName depending on the type of the node.

src/components/Code.tsx
```diff
@@ -207,7 +211,12 @@ function renderLowLightNode(node: lowlight.AST.Unist.Node) {
       ? 1
       : 0;
   if (node.type === 'text') {
-    return <Text>{(node as lowlight.AST.Text).value}</Text>;
+    return (
+      <Text
+        key={createKey(index.toString(), (node as lowlight.AST.Text).value)}>
```
chrisglein, line 216: Is the value of lowlight.AST.Text not unique enough on its own? Why do you need index in there too?

I mention this because as Keith said using index as a key is a bit of a red flag that you don't have a good stable index. In a list that never ever changes order it doesn't matter that much, so this is mostly academic. I believe the fallback behavior likely uses the order in the tree (same as index).

chiaramooney: I thought it would fail on repeated lines of text, but I just tried and it seems to be able to handle that case. Just updated the code.

src/components/Code.tsx
```diff
@@ -196,7 +196,7 @@ const lightStyles = StyleSheet.create({
 });
 
 function createKey(index: string, value: string) {
```
chrisglein, line 198: At this point `createKey` isn't really doing anything and you can remove the function.

chrisglein: Also doesn't appear to be called anymore ;)

chiaramooney: Just updated! Forgot to delete.

src/components/Code.tsx
```diff
@@ -195,7 +195,7 @@ const lightStyles = StyleSheet.create({
   'hljs-attr': {color: lightColors.text}, // Attributes within JSX Tags
 });
 
-function renderLowLightNode(node: lowlight.AST.Unist.Node) {
+function renderLowLightNode(node: lowlight.AST.Unist.Node, index: number) {
```
chrisglein, line 198:
```suggestion
function renderLowLightNode(node: lowlight.AST.Unist.Node) {
```
Unused?

src/components/Code.tsx
```diff
@@ -207,7 +207,11 @@ function renderLowLightNode(node: lowlight.AST.Unist.Node) {
       ? 1
       : 0;
   if (node.type === 'text') {
-    return <Text>{(node as lowlight.AST.Text).value}</Text>;
+    return (
```
chrisglein, line 210: Could have a variable to avoid casting twice below:
```
let text = (node as lowlight.AST.Text);
return (
  <Text key={text}>{text}</Text>
    );

File: src/components/Code.tsx [hunk 7]
```diff
@@ -262,9 +266,9 @@ function renderLowLightNode(node: lowlight.AST.Unist.Node) {
 
 function renderLowLightTree(tree: lowlight.HastNode[]) {
   if (tree.length === 1) {
-    return renderLowLightNode(tree[0]);
+    return renderLowLightNode(tree[0], 0);
   } else {
-    return <>{tree.map(renderLowLightNode)}</>;
+    return <>{tree.map((node, index) => renderLowLightNode(node, index))}</>;
```
chrisglein, line 271: Can revert this line to its previous state if you remove the `index` argument.

chrisglein (approved):

PR #46, opened 2021-01-25
Add page for react-native-pdf component

tritao:
ApplicationFrameHost_4k64aH3t4K


tritao:
CI is failing: `You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 0 MB`

package.json
```diff
@@ -14,6 +14,8 @@
     "@react-native-community/clipboard": "^1.5.1",
     "@react-native-community/datetimepicker": "^3.0.8",
     "@react-native-community/masked-view": "^0.1.10",
+    "@react-native-community/progress-bar-android": "^1.0.4",
+    "@react-native-community/progress-view": "JaneaSystems/progress-view#windows-fix",
```
chiaramooney, line 18: Are these dependencies needed for this PR?

src/RNGalleryList.ts
```diff
@@ -55,6 +56,12 @@ export const RNGalleryList: Array<IRNGalleryExample> = [
     key: 'Image',
     component: ImageExamplePage,
   },
+  {
+
+
```
chiaramooney, line 61: nit: remove extra new lines

package.json
tritao: Yes, not having progress view causes issues.

tritao:
Rebased and addressed issues.

tritao:
Hello @chiaramooney, re-rebased again and updated to your latest example structure. CI is passing as well.

tritao:
Rebased again.

src/examples/PDFExamplePage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <View style={{flex:1, width: state.width, height: state.height}}>
+      <Pdf
+        source={source} 
+        scale={state.scale}
+        horizontal={state.horizontal}
+        style={{flex:1}}/>
+    </View>
+  );
+}`;
```
chiaramooney, line 22: We just made a decision that we want the code excerpts to just be the JSX snippet. Could you remove the import line and in the Pdf component instance use the raw input values rather than the variables? Should look something like this:

'''
<View style={{flex:1, width: 800, height: 800}}>
      <Pdf
        source={{uri:"ms-appx:///dummy.pdf"}} 
        scale={state.scale} // replace with actual data
        horizontal={state.horizontal} // replace with actual data
        style={{flex:1}}/>
    </View>
'''

tritao:
This PR is blocked due to issues with the `rn-fetch-blob` module. which have been appeared since this PR was open.

The latest push updates to the `react-native-pdf` PR at https://github.com/wonday/react-native-pdf/pull/578 which ports to the newer `react-native-blob-util`.

But that is also causing issues, there are compilation errors on Windows, I've opened an issue for it here: https://github.com/RonRadtke/react-native-blob-util/issues/47

jaimecbernardo:
Hi @chiaramooney , 
We've created a fork to fix the `rn-fetch-blob` port build issue and are using that specific commit as a dependency here.
The gallery now builds and this module's page is working on both debug and release.

RonRadtke:
@tritao should be fixed in https://github.com/RonRadtke/react-native-blob-util/releases/tag/0.13.9

chiaramooney (approved):
👍

PR #51, opened 2021-01-29
Set up CI with Azure Pipelines

### LINKED ISSUE #49: CI is failing due to running out of disk space
As the title says, the CI is currently failing on two open PRs due to running out of disk space:

https://github.com/microsoft/react-native-gallery/pull/27
https://github.com/microsoft/react-native-gallery/pull/46

The error: `You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 0 MB`

Github Actions runners only guarantee 14GB of disk space: https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources

A potential solution is probably so setup a Gallery-specific Actions runner (on Azure for instance).

chiaramooney:
Setup pipeline with AzureDevOps using the Azure scale set VM's from react-native-windows in order for native build to have enough disk space. 

Removed Github Actions Pipeline.

Resolves #49


azure-pipelines.yml
```diff
@@ -0,0 +1,72 @@
+# Universal Windows Platform
+# Build a Universal Windows Platform project using Visual Studio.
+# Add steps that test and distribute an app, save build artifacts, and more:
+# https://aka.ms/yaml
```
kmelmon, line 4: Looks good to me, signing off.  @a-coats FYI

kmelmon (approved):
:shipit:

acoates-ms (approved):

PR #63, opened 2021-02-09
Add Badge Indicating Component is a Wrapped Windows Control

### LINKED ISSUE #60: Add indication on pages of whether a control is native or JavaScript
Some React Native controls are implemented directly with their native stack, while some are implemented from primitives in JavaScript and don't heavily lean on larger native controls. It would be helpful to indicate this information somewhere, and the gallery could be a good fit.

There's nothing wrong with being a JavaScript control. So any indicator shouldn't cast that as being a bad thing. It's just additional information for apps that might need to know.

Reasons for wanting to know:
- visual alignment with native controls versus consistency across platforms
- knowledge of behavioral differences that result from the native controls

Ideas for how to indicate status:
- a link to the native docs for the corresponding native control (when there is a clear link)
- some kind of badge

chiaramooney:
Badge added to pages which showcase components that were built by wrapping Windows controls. On these pages, links were also added to the API documentation for the relevant XAML control. So that XAML API link specification is provided alongside "wrapped control" specification, hyperlink section needed to be lifted into Page component. Now all URLs are passed as props into Page. 

Resolves #60  

**_Screenshot of Badge_**
native_control_badge
**_Screenshot of Tooltip Explaining Badge_**
native_control_badge_tooltip


asklar:
It would be better to provide the explanation in a tooltip on the badge, instead of an alert

src/components/NativeControlBadge.tsx
```diff
@@ -0,0 +1,38 @@
...
+    width: 14,
+    resizeMode: 'contain',
+  },
+});
+
+export function NativeControlBadge() {
+  return (
+    <View
+      style={styles.badgeContainer}
+      tooltip="This badge marks that this component has been built by wrapping a native Windows XAML control resulting in the UI of the component matching its native Windows couterpart.">
```
asklar, line 30:
```suggestion
      tooltip="This component wraps a native Windows XAML control: its visual appearance, animations, etc. will always match its native Windows counterpart.">
```

Consider adding a link

asklar, line 30: Consider adding a link, e.g. "For more information, see https://aka.ms/RNW-NativeUI" or something like that

asklar (approved):

kmelmon (approved):
:shipit:

src/examples/CheckBoxExamplePage.tsx
```diff
@@ -91,6 +92,11 @@ function Example() {
             url:
               'https://github.com/react-native-checkbox/react-native-checkbox',
           },
+          {
```
chrisglein, line 95: Since the inclusion of a url and wrappedNativeControl=true are always going to be paired you could make the addition of the badge paired to the existence of the url. As in you would pass in {name: 'CheckBox', url: '[real-url-here]'} as the named parameter to Page, and then page would add both the badge and the link. The advantage of having that single parameter is you can't forget one without the other.

chrisglein (approved):

src/components/Page.tsx
```diff
@@ -8,24 +9,45 @@ const styles = StyleSheet.create({
...
   scrollView: {
     paddingRight: 20,
   },
 });
 
+const DisplayBadge = (wrappedNativeControl: boolean | undefined) => {
+  if (wrappedNativeControl) {
+    return <NativeControlBadge />;
+  } else {
+    return <View />;
```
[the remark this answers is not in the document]
chiaramooney: Updated

src/examples/CheckBoxExamplePage.tsx
chiaramooney: Updated

PR #65, opened 2021-02-11
Fix permissions exception and some other improvements

tritao:
Avoid using HID, Serial and USB permission checks since they can throw UWP exceptions.

So technically the exceptions issue was not really an hard crash, since the RNPermissions code
has a try/catch to deal with them, but under debug mode this causes issues since the default VS
behaviour is to stop the process when the exception is initially thrown.

I've also improved the UI and fixed another place where I was using the wrong permissions key.

Fixes https://github.com/microsoft/react-native-gallery/issues/62.

ApplicationFrameHost_4pwuse1fY7


chiaramooney (approved):
Looks great! Thanks!

PR #67, opened 2021-02-16
Fix Break in Settings Page

chiaramooney:
Upgrade of Page component in #63 broke Settings page. Made updates here to separate Settings page from Page component. UI of Settings page remains identical to before.


kmelmon (approved):
:shipit:

PR #69, opened 2021-02-19
Fix Formatting of Code Snippets

chiaramooney:
Did a pass through all code snippets to update the formatting to match what was discussed in sync-up. 

Formatting Rules :
1. Code snippets should now contain only the JSX snippet for the component. 
2. If possible, functions and function references used should be defined inside of the JSX rather than defined elsewhere and referenced. If this is not possible, do not add the separate function definition to the code snippet. 
3. Do not add state variable declarations to code snippets. If they are referenced in the JSX, that's okay.


kmelmon (approved):
:shipit:

PR #70, opened 2021-02-19
Add Popup Page

chiaramooney:
**_Screenshot of Page_**
image
image
image


src/examples/PopupExamplePage.tsx
```diff
@@ -0,0 +1,319 @@
...
+            setShowPopup2(true);
+          }}>
+          <Text>Open Popup</Text>
+        </TouchableHighlight>
+        <Popup
+          isOpen={showPopup2}
+          isLightDismissEnabled={true}
+          onDismiss={() => {
+            setShowPopup2(false);
+          }}>
```
kmelmon, line 189: Since isLightDismissEnabled == true, do we actually need an onDismiss handler to hide the Popup?

kmelmon (approved):
:shipit:

src/examples/PopupExamplePage.tsx
chiaramooney: I gave it a try without the onDismiss and it seems to glitch because the state variable controlling when to show the pop up is still set to true even though the popup was dismissed. When the user tries to open the popup again, no popup appears.

PR #72, opened 2021-02-23
Add Flyout Page

chiaramooney:
**_Screenshots from Flyout Page_**
image
image
image


src/examples/FlyoutExamplePage.tsx
```diff
@@ -0,0 +1,322 @@
...
+      }}>
+      <Text>Close Flyout</Text>
+    </TouchableHighlight>
+  </View>
+</Flyout>`;
+  const example2jsx = `<TouchableHighlight
+  onPress={() => {
+    setShowFlyout2(true);
+  }}>
+  <Text>Open Popup</Text>
```
kmelmon, line 41: >Popup [](start = 13, length = 5)

I suggest renaming this to "Open Flyout" - as Popup is an implementation detail (think of Flyout as a 1st class popup-like component, rather than a component that has a Popup)

kmelmon (approved):
:shipit:

src/examples/FlyoutExamplePage.tsx
chiaramooney: Updated.

PR #82, opened 2021-03-17
Fix List Warnings for Code Component

chiaramooney:
Previous change revealed errors after merge. Key values collided during the rendering of code snippets. Text value alone is not enough to distinguish between elements in the list because repeated text snippets exist. Lowlight tree does not assign its nodes a static id value to use for identification, so unfortunately index must be used to distinguish list entries with same text contents. Key value is index along with text value to eliminate issue of item at index X being removed and replaced and consequently not re-rendered because the index value alone is constant.


src/components/Code.tsx
```diff
@@ -195,7 +195,15 @@ const lightStyles = StyleSheet.create({
   'hljs-attr': {color: lightColors.text}, // Attributes within JSX Tags
 });
 
-function renderLowLightNode(node: lowlight.AST.Unist.Node) {
+function createTextKey(text: string, index: number) {
```
chrisglein, line 198: Include some comment here about why you're using this as your key. Maybe with some example string so it's clear what the source data looks like.

chrisglein (approved):

PR #89, opened 2021-04-02
Add Workaround To Resolve WebView Touch Scrolling Issue

chiaramooney:
From [this](https://github.com/microsoft/react-native-windows/issues/6287) issue, WebView touch scrolling is broken unless perspective is disabled for the app. Ultimately, the perspective transform for RNW will be altered such that this workaround is no longer needed or a change can be made to the WebView module itself to disable perspective.


windows/rngallery/MainPage.cpp
```diff
@@ -16,5 +16,6 @@ namespace winrt::rngallery::implementation
         InitializeComponent();
         auto app = Application::Current().as<App>();
         ReactRootView().ReactNativeHost(app->Host());
+        ReactRootView().IsPerspectiveEnabled(false);
```
chrisglein, line 19: Include a comment with a link to the RNW issue #?

chiaramooney: Will do!

kmelmon (approved):
:shipit:

windows/rngallery/MainPage.cpp
gangli66: > Will do!

In my react-native-windows project, I add ReactRootView().IsPerspectiveEnabled(false); in two files below:
1. myproject/windows/myproject/MainPage.cpp
2. myproject/node_modules/react-native-windows/template/cpp-app/src/MainPage.cpp
but it still can't scroll inside the webview.

PR #98, opened 2021-04-07
Update Menu Icon

chiaramooney:
Update menu icon to hamburger menu for cleaner look and adjust page layout to fit new dimensions of menu button.

image


src/App.tsx
```diff
@@ -35,7 +47,13 @@ function RNGalleryScreenWrapper({navigation}) {
   const Component = RNGalleryList[state.index].component;
   return (
     <View style={styles.container}>
-      <Button title="Menu" onPress={() => navigation.openDrawer()} />
+      <TouchableHighlight
+        style={styles.menu}
+        onPress={() => navigation.openDrawer()}
+        activeOpacity={0.5783}
```
chrisglein, line 53: Why opacity? Should this be baked into the color? And should it relate to some defined `PlatformColor`?

src/App.tsx
```diff
@@ -35,7 +47,13 @@ function RNGalleryScreenWrapper({navigation}) {
...
+        underlayColor="rgba(0, 0, 0, 0.0241);">
```
chrisglein, line 54: Similar question: Where does this magic opacity number come from?

src/App.tsx
chiaramooney: Sync'd offline

src/App.tsx
chiaramooney: Sync'd offline

chrisglein (approved):

PR #102, opened 2021-04-07
Fix Picker Page

chiaramooney:
A couple improvements for Picker page. Fix up description. Alter picker dimensions to match XAML ComboBox dimensions.


src/examples/PickerExamplePage.tsx
```diff
@@ -37,22 +37,22 @@ export const PickerExamplePage: React.FunctionComponent<{}> = () => {
         },
       ]}>
       <Example title="A simple Picker." code={example1jsx}>
-        <Picker style={{height: 50, width: 100}}>
+        <Picker style={{height: 32, width: 200}}>
```
chrisglein, line 40: This change is fine because it makes the numbers more correct. But later we should reevaluate whether these controls can go without a specified height or not and get the right default.

chrisglein (approved):

src/examples/PickerExamplePage.tsx
chiaramooney: If my memory serves me I believe any time I specified control dimensions in gallery it was because the control wouldn't render unless height/width were specified.

chrisglein: It's certainly been an issue before (see [RNW#5437](https://github.com/microsoft/react-native-windows/issues/5437). But I think @rectified95 addressed some of those and the others we hit should be tracked. It's worth checking in with what controls need and which don't.

rectified95: As far as I remember, the community Picker didn't need dimensions. I may be wrong, though

rectified95:
Side note, @chiaramooney, for a cleaner commit history, you can use rebasing, rather than merging - it will prevent a bunch of noise: 
image

PR #126, opened 2021-04-23
Update React Native Gallery to react-native-windows v0.64.4

chiaramooney:
Update React Native Gallery to react-native-windows v0.64.4


asklar (approved):

windows/rngallery/App.cpp
```diff
@@ -58,7 +57,19 @@ void App::OnLaunched(activation::LaunchActivatedEventArgs const& e)
...
 
     Frame rootFrame = Window::Current().Content().as<Frame>();
-    rootFrame.Navigate(xaml_typename<rngallery::MainPage>(), box_value(e.Arguments()));
+    rootFrame.Navigate(xaml_typename<MainPage>(), box_value(e.Arguments()));
+}
+
+/// <summary>
+/// Invoked when the application is activated by some means other than normal launching.
+/// </summary>
+void App::OnActivated(Activation::IActivatedEventArgs const &e) {
```
AgneLukoseviciute, line 66: out of curiousity - is this required in updating to v0.64.4?

chiaramooney: That function was added automatically by the cli after running `npx react-native-windows-init --version 0.64.4 --overwrite`. The native app template for 0.64.4 must have been updated to add the new method.

AgneLukoseviciute (approved):

windows/rngallery/App.cpp
asklar: technically not needed since the gallery app doesn't do any non-standard activation but we should keep it to stick to the template as much as possible.

PR #129, opened 2021-04-29
Rework the SensitiveInfo example page.

tritao:
This reworks the sensitive info page so the layout makes more sense (swapping the Get and Set item buttons), as well as adding some explanation text to make it clearer how it works.

Fixes https://github.com/microsoft/react-native-gallery/issues/113.

ApplicationFrameHost_BSJLjQKWTM


chiaramooney (approved):
Looks good

PR #133, opened 2021-05-11
Added note on build times

JeanRoca:
Added a note to the README to notify users that the first build time will take longer than the others. This contributions is valuable because it will provide transparency to users that might not be familiar with the extended build times.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ If you wish to build React Native Gallery on your computer locally, follow the f
 2. In the root directory of the repository on your device, run `yarn`.
 3. In the same directory run `npx react-native run-windows`.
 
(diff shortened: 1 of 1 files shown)


stmoy (approved):
:shipit: 

Q: having many modules is common. Does every app suffer like this? Is there anything we can do to make this better?

chrisglein:
> Q: having many modules is common. Does every app suffer like this? Is there anything we can do to make this better?

The pending solution for this is to have the modules all use the canonical Microsoft.ReactNative binaries. Then even if the modules themselves needed to be individually built, they wouldn't all have to also rebuild the core of RNW. That work is tracked in the RNW repo [here](https://github.com/microsoft/react-native-windows/issues/7227). @jonthysell as an FYI

PR #134, opened 2021-05-24
Remove Certificate Password in CI

chiaramooney:
**_What_**
Remove Stale Certificate Password MSBuild prop.

**_Why_**
No longer necessary with certificate from published React Native Gallery as it does not have a password. 

Note: Linter automatically applied other formatting changes.

### DIFF
--- a/ci.yml
+++ b/ci.yml
@@ -1,9 +1,9 @@
 schedules:
-  - cron: "0 0 * * *"
+  - cron: '0 0 * * *'
     displayName: Nightly Build
     branches:
       include:
-      - main
+        - main
     batch: true
 
 variables:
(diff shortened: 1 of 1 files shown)


ci.yml
```diff
@@ -14,67 +14,67 @@ variables:
 pool: windevbuildagents
 
 steps:
-- checkout: self
-  clean: false
+  - checkout: self
```
AgneLukoseviciute, line 17: nit: was the additional whitespace intentional?

AgneLukoseviciute (approved):

ci.yml
rectified95: +1 would be nice if you could remove the indentation changes in this file to make the diff more accurate

chiaramooney: Looks like the linter forced those changes in white space formatting.

PR #135, opened 2021-05-25
Update Hyperlinks to use XAML HyperlinkButton

chiaramooney:
Update Hyperlinks to use XAML HyperlinkButton.


asklar (approved):

asklar:
are there other uses of components/Hyperlink.tsx ? if not, consider removing that file

src/components/LinkTile.tsx
```diff
@@ -27,7 +28,11 @@ export function LinkTile(props: {
     <View style={styles.hyperlinkTile}>
       <Text style={styles.hyperlinkTileTitle}>{props.title}</Text>
       {props.links.map((hyp) => (
-        <Hyperlink key={hyp.label} label={hyp.label} url={hyp.url} />
+        <HyperlinkButton
```
rectified95, line 31: For my information, why are we changing this at all? Is that because `Hyperlink` is a no-op in RNW?
Taking a quick glance at the source code, I can see it's just using text, rather than a native view manager.

rectified95: Oh wait, `Hyperlink` is not a built-in component..
Yeah, so this is def a non-blocking question, just wondering if this is for gallery-scoped use, and we get some benefit from using the native link in terms of styling.

chiaramooney: Yes correct so the Hyperlink component used here was created by me because there was not a com module with windows support for hyperlink when I started gallery. Now that we have react-native-xaml, it's best practice to use its HyperlinkButton because it connects up to native code and has correct Windows color styling for the different states of the button (i.e. hover, pressed, rest, etc).

PR #138, opened 2021-06-01
Update and fix react-native-print page for 0.64

jaimecbernardo:
Updates react-native-print to the latest release, which fixes the module for react-native-windows 0.64.
Re-adds the `RNPrintCanvas` element needed for printing using UWP's APIs.

Fix: https://github.com/microsoft/react-native-gallery/issues/109


chiaramooney (approved):
👍

PR #140, opened 2021-06-02
Add specific commit to sketch-canvas dependency

jaimecbernardo:
Adds a specific commit to the `react-native-sketch-canvas` dependency to try and avoid the need to run `yarn cache clean` when updating this dependency.

Fix: https://github.com/microsoft/react-native-gallery/issues/139


chiaramooney (approved):
👍

PR #143, opened 2021-06-04
Update to latest 0.64 bits

acoates-ms:
Bumping react-native-windows to latest 0.64 bits, and a few other dependencies that were complaining about mismatched peerDeps.


chiaramooney (approved):
👍

PR #145, opened 2021-06-04
Cleanup theme to not use globals

acoates-ms:
Removes the dependency on reactn.
Moves theme management to normal react hooks/contexts.
Fixes logic in Code component to properly get the theme, and stop using the legacy AppTheme module.
Removes some suppressions of typescript errors.


chiaramooney (approved):
Looks great! Thanks

PR #146, opened 2021-06-05
Fix a bunch of lint errors

acoates-ms:
Looks like lint is no longer running correctly.  This fixes the lint command so that lint is running again, and cleans up a bunch of lint warnings.


chiaramooney (approved):
👍

PR #162, opened 2021-09-28
Update Buttons on Permissions Page

chiaramooney:
**_Why_**
Update pressable components on permissions page to use react-native-xaml.


chrisglein:
> **_Why_** Update pressable components on permissions page to use react-native-xaml.

Why is important that these use react-native-xaml? Does this solve a visual styling issue? A side effect however is that the page no longer has the ability to function on Android, iOS, or potential future options like macOS.

chiaramooney:
@chrisglein Ah good point. Was focused on increasing the Windows look and feel of RN Gallery; wasn't thinking about the platform compat issue in moving to react-native-xaml. Will close there PR's now.

PR #164, opened 2021-09-28
Replace Settings Theme Selector with ComboBox

chiaramooney:
**_Why_**
Replace Picker in Settings with react-native-xaml ComboBox


chrisglein:
> **_Why_** Replace Picker in Settings with react-native-xaml ComboBox

How does this affect the behavior and look and feel? Looking at the pages today for Picker vs. the Xaml page that has a ComboBox on it... do they behave identically? (but with different API)

chiaramooney:
Closing for same reason as #162

PR #182, opened 2021-12-29
Test 0.67

chiaramooney:
### Why

Upgrading React Native Gallery to 0.67.

Note:
- Removal of Expander was intentional. Expander currently breaks app. See #183.
- Downgrade of react-native-screens version was intentional. Current version causes screen to go blank when you navigate to a screen that has been visited before. 
Tracking support for screens here: https://github.com/microsoft/react-native-windows/issues/4152


src/RNGalleryList.ts
```diff
@@ -85,12 +85,12 @@ export const RNGalleryList: Array<IRNGalleryExample> = [
     icon: '\uE703',
     type: 'Status and Info',
   },
-  {
+  /*{
```
chrisglein, line 88: Add a comment with the tracking issue # here

windows/rngallery/rngallery.vcxproj
```diff
@@ -6,7 +6,7 @@
     <CppWinRTOptimized>true</CppWinRTOptimized>
     <CppWinRTRootNamespaceAutoMerge>true</CppWinRTRootNamespaceAutoMerge>
     <MinimalCoreWin>true</MinimalCoreWin>
-    <ProjectGuid>{f4d4ef0f-b724-41ed-b6ef-bcf8d91e4775}</ProjectGuid>
+    <ProjectGuid>{5e21afcf-aaf9-4c22-a858-a53bb8cde1fb}</ProjectGuid>
```
chrisglein, line 9: Have we landed on steps to keep the project guid static when we run through a version update?

windows/rngallery/rngallery.vcxproj
```diff
@@ -73,8 +73,11 @@
     <Import Project="..\packages\$(WinUIPackageProps)" Condition="'$(WinUIPackageProps)'!='' And Exists('..\packages\$(WinUIPackageProps)')" />
   </ImportGroup>
   <PropertyGroup Label="UserMacros">
-    <AppxPackageSigningEnabled>False</AppxPackageSigningEnabled>
     <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
+    <AppxPackageSigningEnabled>True</AppxPackageSigningEnabled>
+    <PackageCertificateKeyFile>C:\Users\chiaramooney\GalleryCert\rngallery_Key.pfx</PackageCertificateKeyFile>
```
chrisglein, line 78: I assume this is unintended

metro.config.js
```diff
@@ -4,17 +4,24 @@
...
+
 module.exports = {
   resolver: {
     blockList: exclusionList([
       // This stops "react-native run-windows" from causing the metro server to crash if its already running
       new RegExp(
         `${path.resolve(__dirname, 'windows').replace(/[/\\]/g, '/')}.*`,
       ),
-      // This prevents "react-native run-windows" from hitting: EBUSY: resource busy or locked, open msbuild.ProjectImports.zip
+      // This prevents "react-native run-windows" from hitting: EBUSY: resource busy or locked, open msbuild.ProjectImports.zip or other files produced by msbuild
```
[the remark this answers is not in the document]
chiaramooney: I believe that message is autogenerated from the --overwrite script but yes we should update the template because I believe it should be referring to line 25.

windows/rngallery/rngallery.vcxproj
chiaramooney: For now there's no automatic way to keep the guid the same following a version update. If the developer wished to keep it the same they would have to manual undo the changes of the --overwrite script to restore the original guid information.

jonthysell: It's here, and potentially could be backported to 0.67: https://github.com/microsoft/react-native-windows/pull/9453

jonthysell (approved):

PR #197, opened 2022-02-15
Add Divider Lines to Drawer

### LINKED ISSUE #188: Settings menu is 'pinned' but no visual indicator that it is, resulting in UI that looks broken
I like that the settings menu is 'pinned' in the hamburger menu.
But it is not visually distictive, with either a different background color or a line...
When scrolling the controls list this can make it look like there are rendering issues...
image
Especially since the scrollbar is not clearly indicating where the start and stop area's are...

chiaramooney:
### Why

Add divider lines to drawer to clarify which drawer items are meant to float and which are scrolling. 

Resolves #188

## Screenshots

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -128,6 +128,11 @@ function CustomDrawerContent(props) {
         icon={() => {
           return <Text style={styles.icon}>&#xE80F;</Text>;
         }}
+        style={{
+          borderBottomWidth: 0.5,
+          borderColor: PlatformColor('TextControlForeground'),
+          borderRadius: 0,
+        }}
       />
       <ScrollView style={{}} {...props}>
         {RenderDrawer(props)}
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -54,6 +54,16 @@ const styles = StyleSheet.create({
     backgroundColor: PlatformColor('NavigationViewDefaultPaneBackground'),
     height: '100%',
   },
+  drawerTopDivider: {
```
AgneLukoseviciute, line 57: could we use a single style for both, or do you expect these to diverge down the road?

AgneLukoseviciute (approved):

TatianaKapos (approved):

src/App.tsx
chiaramooney: Already have to diverge because one uses bottom and one used top for border width.

PR #243, opened 2022-07-06
Narrator Improvements

### LINKED ISSUE #224: In Permissions, after invoking ‘Request’ button, keyboard focus is falling on ‘Source code’ link.
Found during accessibility grading. 

Tracking internal bug: https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/40127783

### LINKED ISSUE #226: Narrator is not announcing name of the button, while navigating in the left navigation pane.
Found during accessibility grading.

Tracking internal bug: https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/40128025

### LINKED ISSUE #227: Screen Reader is not announcing the exact state of expand/collapse button in navigation pane.
Found during accessibility grading.

Tracking internal bug: https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/40128113

### LINKED ISSUE #228: Screen Reader is not announcing the group names i.e., Basic Input, Layout, text and etc while entering or existing from the related group of buttons.
Found during accessibility grading.

Tracking internal bug: https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/40128192

### LINKED ISSUE #229: On Permissions page, Screen Reader fails to announce “permissions is not available” text.
Found during accessibility grading.

Tracking internal bug: https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/40128466

TatianaKapos:
## Description
Narrator now calls out the state (expanded or collapse) of navigation bar, the group of controls on home page, and correctly calls out button request/state on Permissions Example Page.

### Why
Resolves #227
Resolves #228
Resolves #224
Resolves #226
Resolves #229

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,7 +11,11 @@ import {
   NavigationContainer,
   useNavigationState,
 } from '@react-navigation/native';
-import {createDrawerNavigator, DrawerItem, getIsDrawerOpenFromState} from '@react-navigation/drawer';
+import {
+  createDrawerNavigator,
+  DrawerItem,
+  getIsDrawerOpenFromState,
+} from '@react-navigation/drawer';
 import RNGalleryList from './RNGalleryList';
 import LightTheme from './themes/LightTheme';
 import DarkTheme from './themes/DarkTheme';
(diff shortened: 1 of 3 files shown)


chiaramooney (approved):
👍

AgneLukoseviciute:
> @AgneLukoseviciute yeah that's a good idea! Which issue were you looking at specifically? Most of these were resolved with just adding an accesssibilityLabel so not sure the expected output when using react-navigation drawer

I think we should be able to add the accessibilityState expanded prop (which fixed #227) to their [DrawerToggleButton](https://github.com/react-navigation/react-navigation/blob/main/packages/drawer/src/views/DrawerToggleButton.tsx) so the added accessibility is provided for "free" and apps don't have to do extra work. 

And then for the one where you added accessibilityLabels (#226), ideally react-navigation would just use the text of the button as the default accessibilityLabel so we don't _need to_ set it explicitly.

PR #267, opened 2022-09-13
Fix UIA Tree

chiaramooney:
### Why

React Native Gallery currently has the issue where the UIA has n copies of the currently mounted page where n is the number of pages that have been navigated to during the app's lifetime. This is incorrect behavior.

### What

Using RNGalleryScreenWrapper to wrap screen upon adding them to the navigation is not legal behavior for the react navigation module. When a new screen is navigated to all screens on the stack are reloaded. Using the current index to determine which screen to wrap, results in the same screen being re-rendered over and over instead of each screen on the stack including the ones not currently loaded. 

Wrapping of screens has now been redone so that a hamburger menu can be displayed on all screens while UIA tree remains correct. This is done by editing the component pages themselves to be wrapped in the correct content.

## Screenshots

No visual changes.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,17 +6,13 @@ import {
   Text,
   useColorScheme,
   ScrollView,
-  Pressable,
 } from 'react-native';
-import {NavigationContainer, useNavigation} from '@react-navigation/native';
+import {NavigationContainer} from '@react-navigation/native';
 import {
   createDrawerNavigator,
   DrawerItem,
   getIsDrawerOpenFromState,
-  DrawerContentScrollView,
-  DrawerItemList,
 } from '@react-navigation/drawer';
-import {createStackNavigator} from '@react-navigation/stack';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -221,16 +152,29 @@ function renderScreen(i: number) {
...
+      {screens}
+    </Drawer.Navigator>
+  );
+}
+
 export default function App(props) {
   const [rawtheme, setRawTheme] = React.useState<ThemeMode>('system');
   const colorScheme = useColorScheme();
   const theme = rawtheme === 'system' ? colorScheme! : rawtheme;
-  appVersion = `${props.MajorVersion}.${props.MinorVersion}.${props.BuildVersion}.${props.RevisionVersion}`;
```
AgneLukoseviciute, line 233: maybe let's add an issue to track the issue with appVersion? otherwise lgtm

AgneLukoseviciute (approved):

PR #269, opened 2022-09-22
Fix Screen Reader failing to announce “permissions is not available” text

### LINKED ISSUE #229: On Permissions page, Screen Reader fails to announce “permissions is not available” text.

TatianaKapos:
## Description
Narrator reads out result from pressing permission button now by adding accessible property to View

### Why
Resolves  #229

### DIFF
--- a/src/examples/PermissionsExamplePage.tsx
+++ b/src/examples/PermissionsExamplePage.tsx
@@ -129,7 +129,7 @@ export const PermissionsExamplePage: React.FunctionComponent<{}> = () => {
         <Text style={{fontWeight: 'bold', paddingLeft: 10, color: colors.text}}>
           {perm}
         </Text>
-        <View focusable={getResultString(status) ? true : false}>
+        <View focusable={getResultString(status) ? true : false} accessible={getResultString(status) ? true : false}>
           <Text
             accessibilityLabel={getResultString(status)}
             style={{paddingLeft: 10, color: colors.text}}>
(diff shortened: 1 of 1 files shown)


src/examples/PermissionsExamplePage.tsx
```diff
@@ -129,7 +129,7 @@ export const PermissionsExamplePage: React.FunctionComponent<{}> = () => {
         <Text style={{fontWeight: 'bold', paddingLeft: 10, color: colors.text}}>
           {perm}
         </Text>
-        <View focusable={getResultString(status) ? true : false}>
+        <View focusable={getResultString(status) ? true : false} accessible={getResultString(status) ? true : false}>
```
AgneLukoseviciute, line 132:
```suggestion
        <View focusable={getResultString(status)} accessible={getResultString(status)}>
```

AgneLukoseviciute, line 132:
```suggestion
        <View focusable={getResultString(status)} accessible={getResultString(status)}>
```

AgneLukoseviciute: I think we can simplify this, but other than that lgtm

AgneLukoseviciute (approved):

PR #317, opened 2023-03-23
Remove Gesture Handler Module with Windows Support

chiaramooney:
### Why

Remove react-native-gesture-handler module with Windows support. Module has not worked, and is prone to native and JS breaks causing reliability issues with RNGallery.


TatianaKapos (approved):
LGTM! :)

Issue #302, opened 2023-02-06
[React Native Gallery -> Popup]: When user invokes 'Open Popup' button, keyboard focus stays on same button and narrator stays mute. 


AgneLukoseviciute:
Repro-Steps:

Launch the React Native Gallery app.
Navigate to ‘Popup’ button and invoke it.
Navigate to 'Open Popup' button and invoke it.
Listen to narrator announcement.
Actual Result:

When user invokes 'Open Popup' button, keyboard focus stays on same button and narrator stays mute. 

Expected Result:

When user invoke 'Open Popup' button, Keyboard focus should move to the opened dialog and narrator should announce about the information about it and while navigating in the dialog focus should stay within the dialog.

internal bug tracked here: https://microsoft.visualstudio.com/OS/_workitems/edit/42926599

AgneLukoseviciute:
The issue is that keyboard focus remains on the "open popup" button. Popup should probably automatically take focus when invoked. We could workaround this by manually forcing the focus, but we should fix this in core (check that it repro's outside of gallery).

chrisglein:
Popup may have an `autoFocus` property, although it's not shown in the docs: https://microsoft.github.io/react-native-windows/docs/popup-component
Is in the view manager: https://github.com/microsoft/react-native-windows/blob/6a6a622535f81ef0f588bade0a7efc5383fc0924/vnext/Microsoft.ReactNative/Views/PopupViewManager.cpp#L172

PR #319, opened 2023-04-06
[React Native Gallery -> Popup]: Automatically shift focus to Popup when user invokes 'Open Popup' button

YajurG:
## Description
Shift the focus to the Popup component automatically when user presses 'Open Popup' button on Popup page in Gallery. 

### Why
Resolves [#302]

### What
When the Popup page is initially opened from the home screen, Narrator focus does not automatically shift to the opened Popup when the 'Open Popup' button is invoked for the first time. Reopening the popup after the initial press will result in focus shifting automatically to the Popup.

## Screenshots

- First press after opening Popup page
Before pressing 'Open Popup':
image
After pressing 'Open Popup' (not focused on Popup):
Screenshot 2023-04-05 172602

-Subsequent presses on the Popup page
Before pressing 'Open Popup':
image
After pressing 'Open Popup' (focus has auto shifted to Popup):
Screenshot 2023-04-05 172827

Note: This happens on the first press immediately after any time the Popup page is loaded from the homepage i.e if you are in the Popup page, go back to the homepage and re-open the Popup page, focus will not shift on the first time you press the 'Open Popup' button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "react-native-tts": "ak1394/react-native-tts",
     "react-native-webview": "^11.18.2",
     "react-native-windows": "^0.70.0",
-    "react-native-xaml": "^0.0.72"
+    "react-native-xaml": "^0.0.74"
   },
   "devDependencies": {
     "@babel/core": "^7.8.4",
--- a/src/examples/PopupExamplePage.tsx
+++ b/src/examples/PopupExamplePage.tsx
@@ -1,6 +1,6 @@
 'use strict';
 import {AccessibilityInfo, Text, TextInput, TouchableHighlight, View, findNodeHandle} from 'react-native';
-import React, {useState, useRef} from 'react';
+import React, {useState, useEffect} from 'react';
 import {Example} from '../components/Example';
 import {Page} from '../components/Page';
 import {Popup} from 'react-native-windows';
@@ -16,8 +16,21 @@ export const PopupExamplePage: React.FunctionComponent<{}> = () => {
 
   const popupRef : React.RefObject<Popup> = React.createRef<Popup>();
   const viewRef: React.RefObject<View> = React.createRef<View>();
-  const textInputRef: React.RefObject<TextInput> = React.createRef<TextInput>();
-  const textRef: React.RefObject<Text> = React.createRef<Text>();
+
+  useEffect(() => {
+    console.log('useEffect called')
+    if (popupRef.current?.props.isOpen) {
+      console.log('popup is open')
+      const reactTag = findNodeHandle(viewRef.current);
+      console.log(reactTag)
+      if (reactTag) { 
+        console.log('setting focus')
+        AccessibilityInfo.setAccessibilityFocus(reactTag);
+      }
+    }
+  })
+
+  
 
   const example1jsx = `<TouchableHighlight
   onPress={() => {
--- a/src/examples/PopupExamplePage.tsx
+++ b/src/examples/PopupExamplePage.tsx
@@ -18,19 +18,13 @@ export const PopupExamplePage: React.FunctionComponent<{}> = () => {
   const viewRef: React.RefObject<View> = React.createRef<View>();
 
   useEffect(() => {
-    console.log('useEffect called')
     if (popupRef.current?.props.isOpen) {
-      console.log('popup is open')
       const reactTag = findNodeHandle(viewRef.current);
-      console.log(reactTag)
-      if (reactTag) { 
-        console.log('setting focus')
+      if (reactTag) {
         AccessibilityInfo.setAccessibilityFocus(reactTag);
       }
     }
   })
-
-  
 
   const example1jsx = `<TouchableHighlight
   onPress={() => {
@@ -172,22 +166,15 @@ export const PopupExamplePage: React.FunctionComponent<{}> = () => {
           }}
           onPress={() => {
             setShowPopup1(true);
-            const reactTag = findNodeHandle(popupRef.current);
-            console.log('onPress')
-            if (reactTag) {
-              console.log('reactTag found')
-              console.log(reactTag)
-              AccessibilityInfo.setAccessibilityFocus(reactTag);
-            }
           }}
           activeOpacity={0.2}
           underlayColor={colors.border}>
           <Text style={{color: colors.text}}>Open Popup</Text>
         </TouchableHighlight>
-        <Popup accessible={true}
+        <Popup accessible
+               focusable
                accessibilityHint={'this is the popup'} 
                isOpen={showPopup1}
-               autoFocus={true}
                ref={popupRef}>
           <View
             style={{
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "react-native-tts": "ak1394/react-native-tts",
     "react-native-webview": "^11.18.2",
     "react-native-windows": "^0.70.0",
-    "react-native-xaml": "^0.0.72"
+    "react-native-xaml": "^0.0.74"
```
AgneLukoseviciute, line 44: was this an intended change?

src/examples/PopupExamplePage.tsx
```diff
@@ -1,6 +1,6 @@
 'use strict';
-import {Text, TouchableHighlight, View} from 'react-native';
-import React, {useState, useRef} from 'react';
+import {AccessibilityInfo, Text, TextInput, TouchableHighlight, View, findNodeHandle} from 'react-native';
```
AgneLukoseviciute, line 2: Do you still need TextInput here?

AgneLukoseviciute (commented):
Looking at XAML I think ideally we'd want the Narrator focus to go directly to the first text in the pop-up rather than the pop-up itself. But as we discussed, with the limitations of setAccessibilityFocus right now that's not possible. I think this is a good workaround! Make sure to open an issue for fixing up setAccessibilityFocus and Popup autofocus.

AgneLukoseviciute:
Just saw this comment:
>>Note: This happens on the first press immediately after any time the Popup page is loaded from the homepage i.e if you are in the Popup page, go back to the homepage and re-open the Popup page, focus will not shift on the first time you press the 'Open Popup' button.

This doesn't sound great, do you know why this is happening? Also does this work for all of the pop-up examples on that sample page? to resolve the bug from accessibility testing we'll definitely need this working on all of the pop-ups we have.

YajurG:
The main issue seems to be that it doesn't automatically shift focus on the first press after the page loads - not sure why. When debugging there was no difference in the code flow between the first press and subsequent presses, all log statements seemed to be hit in the same order. The `setAccessibilityFocus()` is being called as well on the first press, and all properties and state variables are being updated in the same order.

YajurG:
> And can you also confirm that if you keep tabbing, focus isn't going to any of the elements outside of the flyout?

After focusing on the Popup, if you keep tabbing, the focus does not go any elements outside the Popup window - it only resets back to the Popup page after the Popup tab is closed.

chiaramooney:
I agree with @Agne that we'll want to find the culprit of the first press not triggering automatic focus. Without resolution, I don't believe this fix will resolve the filed accessibility grading bug. Did you try adding your SetTimeout code to gallery to see if that would have an effect? 

Also for Example 4 we still want the same focus behavior. The "button as target" is about how the popup should position itself. In this case, it should set its location relative to a button on the screen.

src/examples/PopupExamplePage.tsx
```diff
@@ -14,7 +14,17 @@ export const PopupExamplePage: React.FunctionComponent<{}> = () => {
...
+  const viewRef: React.RefObject<View> = React.createRef<View>();
+
+  useEffect(() => {
+    if (popupRef.current?.props.isOpen) {
+      const reactTag = findNodeHandle(viewRef.current);
+      if (reactTag) {
+        AccessibilityInfo.setAccessibilityFocus(reactTag);
+      }
+    }
+  })
```
chiaramooney, line 27: We'll want to apply this setAccessibilityFocus code to all Popups on the page. So this logic should handle checks to see if those popups are open as well.

YajurG: I'll push the code change that applies the `setAccessibilityFocus` code to all the popup examples.

YajurG:
> I agree with @Agne that we'll want to find the culprit of the first press not triggering automatic focus. Without resolution, I don't believe this fix will resolve the filed accessibility grading bug. Did you try adding your SetTimeout code to gallery to see if that would have an effect?

Using setTimeout with 100 as the delay works - I double checked and rebuilt gallery and tried to make sure :). I had initially tried with a shorter delay which made no change, but the 100 delay focuses all four of the example popups automatically. The delay also doesn't appear noticeable when the focus auto-shifts.

> Also for Example 4 we still want the same focus behavior. The "button as target" is about how the popup should position itself. In this case, it should set its location relative to a button on the screen.

Makes sense - the current code has the focus automatically shifting to the Popup for Example 4.

chiaramooney:
@Yajur Sounds good! Did you get a chance to check in with Andrew or Jon to make sure the usage of the timer doesn't cause any particulary bad side effect? We'll want to get confirmation from one of them before we merge this.

YajurG:
> @yajur Sounds good! Did you get a chance to check in with Andrew or Jon to make sure the usage of the timer doesn't cause any particularly bad side effect? We'll want to get confirmation from one of them before we merge this.

Yes - he said that I could just use the `focus()` API instead of `setAccessibilityFocus` since `focus` will also shift accessibility focus and work with narrator - made those changes and changed the timeout to 50 which still worked with all four of the Popup examples

AgneLukoseviciute:
> > @yajur Sounds good! Did you get a chance to check in with Andrew or Jon to make sure the usage of the timer doesn't cause any particularly bad side effect? We'll want to get confirmation from one of them before we merge this.
> 
> Yes - talked to Andrew and he said that I could just use the `focus()` API instead of `setAccessibilityFocus` since `focus` will also shift accessibility focus and work with narrator - made those changes and changed the timeout to 50 which still worked with all four of the Popup examples

can you try if focus() works on text? Ideally we match WinUI behavior and focus on the first text element within the Popup rather than the view of the popup itself.

src/examples/PopupExamplePage.tsx
```diff
@@ -14,7 +21,49 @@ export const PopupExamplePage: React.FunctionComponent<{}> = () => {
...
+  const view3Ref: React.RefObject<View> = React.createRef<View>();
+  const view4Ref: React.RefObject<View> = React.createRef<View>();
+
+  useEffect(() => {
+    setTimeout(setFocus, 50);
+  });
+
+  const setFocus = () => {
+    if (popupRef.current?.props.isOpen) {
+      const reactTag = findNodeHandle(viewRef.current);
```
acoates-ms, line 40: Can we simplify some more and stop calling findNodeHandle?   That method is going away in fabric, and I'm not sure why you need to use it here.

YajurG:
> can you try if focus() works on text? Ideally we match WinUI behavior and focus on the first text element within the Popup rather than the view of the popup itself.

I tried setting focus() on refs for the text components in the Popup, and Gallery would automatically shut down whenever I hit the 'Open Popup' button - I'm guessing because I was trying to focus on a text. Not sure why this is happening but I can create an issue with a code sample that I used showing the behavior.

src/examples/PopupExamplePage.tsx
YajurG: Removed the `findNodeHandle` calls

src/examples/PopupExamplePage.tsx
```diff
@@ -1,6 +1,13 @@
 'use strict';
-import {Text, TouchableHighlight, View} from 'react-native';
-import React, {useState, useRef} from 'react';
+import {
+  AccessibilityInfo,
```
AgneLukoseviciute, line 3: nit: remove the unused AccessibilityInfo, TextInput, and findNodeModuleHandle imports

AgneLukoseviciute (approved):

Issue #305, opened 2023-02-27
[React Native Gallery-> Expander]: Visual looking heading “Expander” is not defined as heading.


AgneLukoseviciute:
**Repro-Steps:**
Launch the React Native Gallery app.
Navigate to any button (E.g., Expander) and invoke it.
Set narrator to headings mode using caps lock + Page up/down.
Now navigate to “Expander” heading and listen to the narrator’s announcement.
**Actual Result:**
Visual looking heading “Expander” is not defined as heading. When user navigates to this heading narrator simply announcing as static text like “Expander”.

Note: Same issue is observed in entire application.

**Expected Result:**
Visual heading “Expander” should be defined as the heading. When user navigates to this heading narrator should announce as ““Expander, heading level 1”.

**User Impact:**
Users with visual impairments who rely on screen reader may use heading navigation mode to render through the page. Not defining heading in the application will make it difficult for users navigating in the app.

Internal bug link: https://dev.azure.com/microsoft/OS/_workitems/edit/43361294

AgneLukoseviciute:
We added support for headings in August (https://github.com/microsoft/react-native-gallery/issues/306).

PR #326, opened 2023-05-18
Add header accessibilityRole to Expander heading

YajurG:
## Description

Adds the header accessibilityRole prop to the heading on the Expander page

### Why

Resolves [#305]

### What

- Bumped versions to include changes made in this PR: https://github.com/microsoft/react-native-windows/pull/11603
- Added accessibilityRole prop to the heading component


JunielKatarn (approved):

src/examples/ExpanderExamplePage.tsx
```diff
@@ -72,7 +72,9 @@ export const ExpanderExamplePage: React.FunctionComponent<{}> = () => {
       ]}>
       <Example title="A simple Expander." code={example1jsx}>
         <Expander collapsedHeight={50} expandedHeight={125}>
-          <Text>Text in the header</Text>
+          <Text accessible accessibilityRole="button">
```
AgneLukoseviciute, line 75: was setting accessibilityRole here intentional?

jonthysell (approved):

src/examples/PopupExamplePage.tsx
```diff
@@ -171,6 +171,7 @@ export const PopupExamplePage: React.FunctionComponent<{}> = () => {
             'https://github.com/microsoft/react-native-windows/blob/master/vnext/Microsoft.ReactNative/Views/PopupViewManager.h',
         },
       ]}>
+      <Text accessibilityRole="header">Hello</Text>
```
AgneLukoseviciute, line 174: what's up with this?

src/components/Page.tsx
```diff
@@ -94,7 +94,14 @@ export function Page(props: {
   return isScreenFocused ? (
     <ScreenWrapper style={styles.container}>
       <View style={styles.titlePane}>
-        <Text style={[styles.title, {color: colors.text}]}>{props.title}</Text>
+        <View>
```
AgneLukoseviciute, line 97: do you need the extra view component here?

src/examples/PopupExamplePage.tsx
YajurG: I think this was a test line - removed

src/components/Page.tsx
YajurG: Don't think so - edited

src/examples/ExpanderExamplePage.tsx
YajurG: Actually after looking at it again I don't think we need an accessibilityRole here since it's not the actual heading of the page

PR #333, opened 2023-06-27
Added header accessibilityRole prop to headings

YajurG:
## Description
Added header accessibilityRole prop to headings throughout Gallery

### Why

Resolves https://dev.azure.com/microsoft/OS/_workitems/edit/43361294

### What

Added `accessibilityRole={'header'}` to all headings and sub-headings in Gallery example pages

### DIFF
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -215,6 +215,7 @@ exports[`Button Example Page 1`] = `
       <View>
         <View>
           <Text
+            accessibilityRole="header"
             style={
               {
                 "color": "rgb(28, 28, 30)",
@@ -363,6 +364,7 @@ exports[`Button Example Page 1`] = `
         </View>
         <View>
           <Text
+            accessibilityRole="header"
             style={
               {
                 "color": "rgb(28, 28, 30)",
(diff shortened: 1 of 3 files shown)


AgneLukoseviciute (approved):

AgneLukoseviciute:
A superstar follow-up would be to add this to the create-react app template as well for the heading(s) there.

__tests__/__snapshots__/App-test.js.snap
```diff
@@ -215,6 +215,7 @@ exports[`Button Example Page 1`] = `
       <View>
         <View>
           <Text
```
chrisglein, line 217: Not necessary for this change being made on a timeline, but given how much this is duplicated I think a styled helper component is in order:

```js
<Header text="A simple Button."/>
```

And somewhere else co-defined:
```js
function Header({text: string}) {
  return (
    <Text            
      accessibilityRole="header"
      style={
        {
          "color": "rgb(28, 28, 30)",
          "fontSize": 20,
          "marginBottom": 10,
          "marginTop": 30,
        }
      }
    >
      {text}
    </Text>
  );
}
```

chrisglein (approved):

PR #334, opened 2023-06-27
Fixing null Name property on UIA tree

YajurG:
## Description
Added accessibilityLabel to component examples where UIA name property was null
### Why

Resolves https://dev.azure.com/microsoft/OS/_workitems/edit/43367679

### What

Added the accessibilityLabel prop to component example pages in Gallery where the name property on the UIA tree for components was null.

### DIFF
--- a/src/components/Example.tsx
+++ b/src/components/Example.tsx
@@ -35,7 +35,9 @@ export function Example(props: {
   const styles = createStyles(colors);
   return (
     <View>
-      <Text accessibilityRole={'header'} style={styles.title}>{props.title}</Text>
+      <Text accessibilityRole={'header'} style={styles.title}>
+        {props.title}
+      </Text>
       {props.code ? (
         <View style={styles.box}>
           <View style={styles.exampleContainer}>{props.children}</View>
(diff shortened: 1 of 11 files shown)


src/components/Example.tsx
```diff
@@ -35,7 +35,9 @@ export function Example(props: {
   const styles = createStyles(colors);
   return (
     <View>
-      <Text style={styles.title}>{props.title}</Text>
+      <Text accessibilityRole={'header'} style={styles.title}>
```
AgneLukoseviciute, line 38: same here

src/examples/CheckBoxExamplePage.tsx
```diff
@@ -25,38 +25,57 @@ export const CheckBoxExamplePage: React.FunctionComponent<{}> = () => {
         },
       ]}>
       <Example title="A simple Checkbox." code={'<CheckBox />'}>
-        <CheckBox />
+        <CheckBox accessibilityLabel="Simple Checkbox" />
```
AgneLukoseviciute, line 28: https://learn.microsoft.com/en-us/windows/win32/winauto/uianameshouldnotcontaincontroltype you'll likely get hit with some of these errors. Try changing it to something like "simple example" "checked example" etc. 

Same for all of the other components.

AgneLukoseviciute (approved):

Issue #337, opened 2023-06-29
[React Native Gallery-> Expander]: Visual looking headings are not defined as heading in Home and Settings pages.


YajurG:
**Pre-Requisite:**
Turn on Narrator (Ctrl + Win + Enter).

**Repro-Steps:**

1. Launch the React Native Gallery app.
2. Navigate to Home Page and invoke it.
3. Set narrator to headings mode using caps lock + Page up/down.
4. Observe different bold texts are not defined as heading
5. Navigate to Settings Page and observe the same there too
6. Navigate any other page and observe the heading for:
        - View page code at github
        - Documentation 
        - Feedback

**Actual Result:**
Visual looking headings are not defined as heading. When user navigates to these heading narrator simply announcing as static text.

**Expected Result:**
Visual headings should be defined as the heading programmatically as well. 

**User Impact:**
Users with visual impairments who rely on screen reader may use heading navigation mode to render through the page. Not defining heading in the application will make it difficult for users navigating in the app.

Internal bug link: https://microsoft.visualstudio.com/OS/_workitems/edit/45265286

PR #340, opened 2023-06-29
Changed button color on Button page to be visible in High Contrast theme modes

### LINKED ISSUE #338: [React Native Gallery -> Flyout]: The text inside the colored button is not visible properly in high contrast themes.
**Pre-Requisite:**  

Turn on High Contrast Aquatic Theme (Settings->Accessibility->Contrast themes->Aquatic/Dessert theme->Apply)

**Repro-Steps:**

1. Launch the React Native Gallery app.
2. Navigate to ‘Button’ button and invoke it.
3. Observe Colored button is not having proper contrast in high contrast mode

**Actual Result:**

The text inside the button is not clearly visible in high contrast themes.

**Expected Result:**

All the content should be clearly visible in high contrast themes. 

**User impact:**

Users with low vision will not be able to consume the content present in the Flyout card if it is not visible in High Contrast Themes.

Internal bug link: https://microsoft.visualstudio.com/OS/_workitems/edit/45265387

YajurG:
## Description
Changed button color on Button example page  in High Contrast theme modes

### Why

Resolves #338 

### What

Used Platform Color API to change color of button so that it is visible in both regular and high-contrast theme modes.

## Screenshots
Normal: 
image

Aquatic theme: 
image


YajurG:
This contains commits from other issue as well - checked out from issue #337 branch by mistake.

.github/policies/resourceManagement.yml
marlenecota, line 1: what is this change? linting or something? Doesn't look like it should be part of this PR.

src/HomePage.tsx
marlenecota, line 1: Remove this if this is being fixed in a different PR.

src/SettingsPage.tsx
marlenecota, line 1: Remove this if this is being fixed in a different PR.

__tests__/__snapshots__/App-test.js.snap
```diff
@@ -445,7 +445,7 @@ exports[`Button Example Page 1`] = `
                           "textAlign": "center",
                         },
                         {
-                          "color": "rgb(0, 122, 255)",
+                          "color": "silver",
```
marlenecota, line 448: If the test is passing, does this mean that "SystemColorButtonFaceColor" is a valid windowsbrush?

marlenecota (changes requested):

src/examples/ButtonExamplePage.tsx
```diff
@@ -48,7 +48,11 @@ export const ButtonExamplePage: React.FunctionComponent<{}> = () => {
       <Example title="A colored Button." code={example2jsx}>
         <Button
           title="Button"
-          color={colors.primary}
+          color={
+            Platform.OS === 'windows'
+              ? PlatformColor('SystemColorButtonFaceColor')
+              : 'silver'
+          }
```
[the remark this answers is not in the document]
YajurG: So I think one reason we wanted to use PlatformColor here is because of some accessibility guidelines when in high-contrast mode which are easier to resolve with certain PlatformColor colors - i.e they want the color of the text to contrast with the color of the background in both normal and high-contrast theme modes (if the background is dark, the text should be light and vice versa).

Although the current PlatformColor also resolves to the same color as 'silver', one argument for keeping the PlatformColor usage vs hard coding a color is that hard coding doesn't take the background color into consideration, which can lead to light text on a light background or similar situations where there is low contrast which leads to lower visibility ([link](https://learn.microsoft.com/en-us/windows/apps/design/accessibility/high-contrast-themes#hard-coded-colors) for reference).

__tests__/__snapshots__/App-test.js.snap
YajurG: Yup - I think it is resulting in the same color as 'silver'

YajurG:
I used the rebase command to remove the unwanted commits - think the relevant changes for the PR are now present.

TatianaKapos (approved):

marlenecota (approved):

PR #343, opened 2023-07-05
Bump gallery to RNW .72

TatianaKapos:
## Description
Bumps Gallery to RNW .72

### What
Due to inactive Community Modules, we had to patch the following CMS to the right WindowsTargetPlatformMinVersion. We
can remove the patches when they take in the changes listed here (https://github.com/microsoft/react-native-gallery/issues/324) and publish a new version.

- RNTrackPlayer -Currently use a fork of the main repository (we should figure out why but example was added by 3rd-party) which hasn't been updated since 2021
- RNSenstivieInfoCPP - haven't updated since 2021
- SliderWindows - new version is not out yet
- Clipboard -new version is not out yet
- RNTSS -haven't updated since 2021
- RNCConfig - PR not reviewed yet
- Progress-view - new version is not out yet
- RNSketchCanvas - haven't updated since 2022 and needs to update from Win2D.uwp.1.25.0 to 1.26.0 to match Expander
-  RNPrint - PR not reviewed yet

Created follow up issue to deal with Flyout and Popout jest tests: https://github.com/microsoft/react-native-gallery/issues/344


YajurG (approved):

.github/policies/resourceManagement.yml
chiaramooney, line 1: What's the reason for this file being deleted?

package.json
```diff
@@ -8,54 +8,57 @@
...
-    "@react-native-community/datetimepicker": "^6.1.3",
+    "@react-native-community/checkbox": "^0.5.16",
+    "@react-native-community/datetimepicker": "^7.3.0",
     "@react-native-community/masked-view": "^0.1.11",
-    "@react-native-community/progress-view": "^1.4.0",
-    "@react-native-community/slider": "^4.0.0-rc.3",
-    "@react-native-picker/picker": "^2.4.4",
+    "@react-native-community/progress-view": "^1.4.1",
+    "@react-native-community/slider": "^4.4.2",
+    "@react-native-picker/picker": "^ 2.4.10",
```
chiaramooney, line 22: nit: extra space

windows/rngallery/.gitignore
```diff
@@ -1,2 +1 @@
 /Bundle
-*.pfx
```
chiaramooney, line 2: We'll want to add this line back in. It's a safe guard against publishing pfx files to the repo.

windows/rngallery/MainPage.cpp
```diff
@@ -17,19 +16,5 @@ namespace winrt::rngallery::implementation
         InitializeComponent();
         auto app = Application::Current().as<App>();
         ReactRootView().ReactNativeHost(app->Host());
-        // Workaround for https://github.com/microsoft/react-native-windows/issues/6287
```
chiaramooney, line 20: We'll want to add this code back in. Needed for a webview workaround and versioning.

windows/rngallery/MainPage.xaml
```diff
@@ -8,12 +8,9 @@
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     mc:Ignorable="d"
     Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
-    <Grid>
```
chiaramooney, line 11: We'll want to add this back in. Needed for Print module.

.github/policies/resourceManagement.yml
chrisglein: This was added by the move from fabricbot to the new bot infrastructure. It contains the logic the bot uses for applying triage labels and such. It shouldn't be affected by any code changes.

patches/@react-native-clipboard+clipboard+1.11.2.patch
chrisglein, line 1: The meaningful change in this patch is small. Like one line, updating the min version. But there's a ton of extra content in the patch file. I've hit this too, and as far as I can tell it's getting some of the generated code in the patch file, which isn't quite right. I've started manually editing the patch files to make sure they stay minimal, although it's a bit of a pain. I've also had lunch making the code change, running patch package before I build, and then it's only the code changes I expect and not the generated files too.

TatianaKapos: good call out! Didn't realize how large those files were, changed them to only include the small file change :)

chiaramooney (approved):
Looks good to me!

PR #345, opened 2023-07-07
Update bot yml to match version in samples repo

chrisglein:
## Description

Updates the yml configuration that our bot uses for labeling/lifetime/etc.. Specifically, this takes the copy from [react-native-windows-samples](https://github.com/microsoft/react-native-windows-samples/blob/main/.github/policies/resourceManagement.yml) and copies it to Gallery.

### Why

There's not intended to be a difference between them.
Also there's been some weirdness in PRs touching this file recently and maybe it had some whitespace changes that needed a refresh?

### Other

- There are some capitalization differences in the labels. Not sure if case insensitive. But I can also just edit the labels to match (which is a good idea anyway).
- This brought in the "needs environment info" rule, which seems like goodness. May need to add the corresponding label.

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -16,7 +16,7 @@ configuration:
       - isIssue
       - isOpen
       - hasLabel:
-          label: 'Needs: author feedback'
+          label: 'Needs: Author Feedback'
       - hasLabel:
           label: 'Status: no recent activity'
       - noActivitySince:
@@ -31,7 +31,7 @@ configuration:
       - isIssue
       - isOpen
       - hasLabel:
-          label: 'Needs: author feedback'
+          label: 'Needs: Author Feedback'
       - noActivitySince:
           days: 7
       - isNotLabeledWith:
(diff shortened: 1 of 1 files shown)


chiaramooney (approved):
👍

YajurG (approved):
Since I couldn't pull the change for bumping Gallery to 0.72, I just re-cloned the Gallery repo and the `resourceManagement.yml` 'change' disappeared from my unstaged changes because I think it was a part of the PR that Tatiana merged in when bumping 0.72, so that may be all that's needed to resolve the issue.

chrisglein:
> Since I couldn't pull the change for bumping Gallery to 0.72, I just re-cloned the Gallery repo and the `resourceManagement.yml` 'change' disappeared from my unstaged changes because I think it was a part of the PR that Tatiana merged in when bumping 0.72, so that may be all that's needed to resolve the issue.

Yeah, looks like @TatianaKapos took a change to the file in the version bump. Whitespace only changes by the look of it? I'll merge that and see if we're in a good state. Still good to start matching these files across our repos now that they're text comparable.

PR #347, opened 2023-07-11
React Native Lottie Animation Integration Committe

Victor0814gui:
## Description
adding example page for lottie-react-native https://github.com/microsoft/react-native-gallery/issues/341

### What

since the lottie-react-native package doesn't connect with the list of package managers, I added them directly to the App.cpp file.

rngallery.vcxproj modified to add another lottie-react-native and anm path reference


Victor0814gui:
@microsoft-github-policy-service agree

TatianaKapos:
Hi, thanks for the contribution! Would you mind adding a screenshot of the example page?

Victor0814gui:
I completely forgot, sorry.
Captura de Tela (523)

Directory.Build.props
```diff
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
-
+   <PropertyGroup>
```
chiaramooney, line 3: This shouldn't be needed. The latest version of gallery should already be using WinUI2.8.

src/examples/LottieAnimationsExamplePage.tsx
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  useEffect(() => {
+    anim.current?.play();
+  }, []);
+
+  return (
+    <Page
+      title="Lottie animations"
+      description="Lottie is a library ecosystem for parsing animations from Adobe After Effects exported as JSON with Bodymovin and rendering them natively!."
```
chiaramooney, line 67: nit: Don't need both ! and . at the end of the description

windows/rngallery/rngallery.vcxproj
```diff
@@ -215,4 +233,10 @@
     <Error Condition="!Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppApp.props')" Text="$([System.String]::Format('$(ErrorText)', '$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppApp.props'))" />
     <Error Condition="!Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppApp.targets')" Text="$([System.String]::Format('$(ErrorText)', '$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppApp.targets'))" />
   </Target>
-</Project>
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
+    <PropertyGroup>
+      <ErrorText>Este projeto faz referência a pacotes do NuGet que não estão presentes neste computador. Use a Restauração de Pacotes do NuGet para baixá-los.  Para obter mais informações, consulte http://go.microsoft.com/fwlink/?LinkID=322105. O arquivo ausente é {0}.</ErrorText>
```
chiaramooney, line 238: nit: Translate error text to English

chiaramooney:
Thanks for your contribution! Just a couple of quick comments to address and we can get your page merged in.

Directory.Build.props
Victor0814gui: > This shouldn't be needed. The latest version of gallery should already be using WinUI2.8.

By default, lottie-react-native uses WinUI 2.6, so to override this, they recommend setting that property in the Directory.Build.props file.

chiaramooney: Hmm since the app uses WinUI2.8 I would expect that to take precedence over lottie. Can you confirm that the build doesn't work without this change? If that doesn't work, could you also try specifying WinUI 2.8.0 in the app's ExperimentalFeatures file. See here for info on that: https://github.com/microsoft/react-native-gallery/blob/c7e197a8b6b39733986fd9f3f993dfdc2a0492ec/windows/ExperimentalFeatures.props#L35

Victor0814gui: I have recompiled the project a few times to test, and I have not had any problems with the compilation of the package or the project without the property <WinUI2xVersion>2.8.0</WinUI2xVersion>.

chiaramooney (approved):

PR #357, opened 2023-09-18
Add accessibilityLabel support to DateTimePicker example page

### LINKED ISSUE #336: [React Native Gallery]: The Name property of a focusable element must not be null.
**Repro-Steps:**
Navigate to ‘DatePicker’ button using tab key and invoke it.
Now run fastpass and observe. 

**Actual Result:**
The Name property of a focusable element must not be null. 
 
**Expected Result:**
The Name property of a focusable element must be defined. 

**User impact:**
Users with visual impairments who rely on screen reader will not be able to know the name and purpose of the button if they is not proper name is defined. 

Internal bug link: https://microsoft.visualstudio.com/OS/_workitems/edit/45264469

YajurG:
## Description

Updates the datetimepicker version to include changes for accessibilityLabel support.

### Why

Resolves #336 

### What

Update datetimepicker dependency version to include changes to add accessibilityLabel support: [accesibilityLabel PR](https://github.com/react-native-datetimepicker/datetimepicker/pull/805)

## Screenshots

Before: 
image

After:
image

### DIFF
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -5170,7 +5170,7 @@ exports[`Expander Example Page 1`] = `
                       undefined,
                     ]
                   }
-                  timeZoneOffsetInSeconds={0}
+                  timeZoneOffsetInSeconds={-0}
                 />
               </ExpanderView>
             </View>
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -5170,7 +5170,7 @@ exports[`Expander Example Page 1`] = `
                       undefined,
                     ]
                   }
-                  timeZoneOffsetInSeconds={-0}
+                  timeZoneOffsetInSeconds={-28800}
                 />
               </ExpanderView>
             </View>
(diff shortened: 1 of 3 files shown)


TatianaKapos (approved):

__tests__/__snapshots__/App-test.js.snap
```diff
@@ -5170,7 +5170,7 @@ exports[`Expander Example Page 1`] = `
                       undefined,
                     ]
                   }
-                  timeZoneOffsetInSeconds={-0}
+                  timeZoneOffsetInSeconds={-28800}
```
chrisglein, line 5173: What about this change changed the time zone? 🤣

YajurG: This has been happening on a few of my previous gallery PRs, talked to @chiaramooney about it and there seemed to be a difference in the timezone between the CI checks and the local tests - just need to manually change the value in the snapshot file locally.

chrisglein: @chiaramooney is it possible to omit certain properties from the snapshots?

chiaramooney: To my knowledge, no there isn't. @YajurG if you want to give a quick skim of the Jest Snapshot test docs to confirm that could be good. But yes there does seem to be a mismatch between our local time settings and the CI machines.

PR #372, opened 2023-11-27
Bump RNW to 0.72.20 to include fmt change to fix VS Build errors

YajurG:
## Description

Bump RNW version to ^0.72.0 to include change to fix build errors on VS 2022 v17.8: https://github.com/microsoft/react-native-windows/pull/12412

### What

Bump RNW version to ^0.72.0.


marlenecota:
@chiaramooney - double checking that it would be ok to always use the latest revision of the current rnw version

chiaramooney:
@marlenecota Should be good as long as it builds locally and passes CI!

marlenecota (approved):

PR #373, opened 2023-11-28
Fix 'Disabled Pressable' example on Pressable page

### LINKED ISSUE #363: [React Native Gallery -> Pressable]: Action is defined for disabled button i.e., for 'Disabled Pressable' button present under "A disabled Pressable component".
**Repro-Steps:**

- Launch the React Native Gallery app.
- Navigate to ‘Pressable’ button and invoke it.
- Navigate to 'Disabled Pressable' button, invoke it and observe.

**Actual Result:**

Action is defined for disabled button. On invoking 'Disabled Pressable' button action is being performed, which is incorrect.

**Expected Result:**

For disabled buttons action should be defined. When user navigate and invoke 'Disabled Pressable' button, no action should be performed. 

**User Impact:**

Users with motor disabilities who rely on keyboard will be impacted and misguided if some action is defined for disabled buttons.

Internal Bug Link: https://dev.azure.com/microsoft/OS/_workitems/edit/47843510

YajurG:
## Description
Change Disabled pressable example so that it no longer performs an action.

### Why

Resolves #363 

### What

Removed different display messages when Pressable was being pressed.

## Screenshots
Before: 
Screenshot (35)

After:
Screenshot (34)

### DIFF
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -19364,28 +19364,7 @@ exports[`Pressable Example Page 1`] = `
                   </Text>
                   <Text>
                     }&gt;
-  {
-                  </Text>
-                  <Text
-                    style={{}}
-                  >
-                    <Text>
-                      (
-                    </Text>
-                    <Text
-                      style={{}}
-                    >
-                      <Text>
-                        {pressed}
-                      </Text>
-                    </Text>
-                    <Text>
-                      ) =&gt;
-                    </Text>
-                  </Text>
-                  <Text>
-                     (
-    
+  
                   </Text>
                   <Text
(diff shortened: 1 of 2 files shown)


YajurG:
From the behaviour before, it looks like this may be a bug in pressable, as the text when pressed should not show up when the disabled prop is set to `true`. I have temporarily modified the example, but not sure if I should instead leave as is and open a tracking issue on the RNW repo.

YajurG:
> From the behaviour before, it looks like this may be a bug in pressable, as the text when pressed should not show up when the disabled prop is set to `true`. I have temporarily modified the example, but not sure if I should instead leave as is and open a tracking issue on the RNW repo.

Closing to leave example page as is.

PR #379, opened 2023-12-08
Upgrade to RNW 73

TatianaKapos:
## Description
Upgrade to RNW 73

### Why
Part of release tasklist

### What

Had to manually add @react-native/metro-config to get metro to run. Error message is descriptive enough to solve this issue but wondering if this is specific to upgrading Gallery or if this will affect other apps (this does not effect building a new app).


jonthysell (approved):

chiaramooney (approved):
👍

PR #409, opened 2024-04-17
Add clipboard example page

chrisglein:
### Why

Clipboard is a module already included by Gallery, but doesn't have a sample page. Furthermore, the WinUI gallery has a sample page for clipboard.

### What

Adds a sample page for Clipboard, which behaves identically to the one in WinUI gallery.

## Screenshots

image

### DIFF
--- a/src/RNGalleryList.ts
+++ b/src/RNGalleryList.ts
@@ -44,6 +44,7 @@ let RNGalleryCategories = [
   'Layout',
   'Status and Info',
   'Media',
+  'Text',
   'System',
 ];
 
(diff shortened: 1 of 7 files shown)


src/RNGalleryList.ts
```diff
@@ -36,6 +37,16 @@ import {WindowsHelloExamplePage} from './examples/WindowsHelloExamplePage';
 import {ExpanderExamplePage} from './examples/ExpanderExamplePage';
 import {VirtualizedListExamplePage} from './examples/VirtualizedListExamplePage';
 
+let RNGalleryCategories = [
```
marlenecota, line 40: Did you rename "Text" to "System"?

chrisglein: Not intentionally. "System" was supposed to be an addition. But something got messed in the merge from my other branch and "Text" got replaced. Will fix.

PR #411, opened 2024-04-19
Add radio buttons example

chrisglein:
### Why

Radio buttons are a common UX primitive on desktop, and "how do I create radio groups with RNW" is a question that comes up.

### What

Adds 2 examples that demonstrates how to use radio buttons:
- Sample that uses the `react-native-radio-buttons-group` community module
- Add an example to the `react-native-xaml` page that shows how to use XAML's radio button types

## Screenshots

image

image

### DIFF
--- a/__tests__/App-test.js
+++ b/__tests__/App-test.js
@@ -34,6 +34,7 @@ import {XamlExamplePage} from '../src/examples/XamlExamplePage';
 //import {TrackPlayerExamplePage} from '../src/examples/TrackPlayerExamplePage';
 import {WindowsHelloExamplePage} from '../src/examples/WindowsHelloExamplePage';
 import {ExpanderExamplePage} from '../src/examples/ExpanderExamplePage';
+// https://github.com/microsoft/react-native-gallery/issues/420
 // Unable to import: {export * from './types'; SyntaxError: Unexpected token 'export'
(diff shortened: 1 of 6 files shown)


__tests__/App-test.js
```diff
@@ -34,6 +34,8 @@ import {XamlExamplePage} from '../src/examples/XamlExamplePage';
 //import {TrackPlayerExamplePage} from '../src/examples/TrackPlayerExamplePage';
 import {WindowsHelloExamplePage} from '../src/examples/WindowsHelloExamplePage';
 import {ExpanderExamplePage} from '../src/examples/ExpanderExamplePage';
+// Unable to import: {export * from './types'; SyntaxError: Unexpected token 'export'
+//import {RadioButtonsExamplePage} from '../src/examples/RadioButtonsExamplePage';
```
chiaramooney, line 38: Do you want to leave this test in here? If so, let's file an issue and add it here to mark fixing this later.

chiaramooney (approved):

__tests__/App-test.js
chrisglein: Yep, let me file an issue to track enabling this test. It's probably possible with a test config setting, but I felt out of my depth on that.

PR #412, opened 2024-04-19
Fix badges in dark mode

chrisglein:
### Why

Badges displayed incorrectly in dark mode.

In light mode:
image
In dark mode:
image

### What

Moves to `PlatformColor` styles from XAML, which support light/dark mode.

## Screenshots

image
image

### DIFF
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -1295,7 +1295,7 @@ exports[`CheckBox Example Page 1`] = `
               {
                 "backgroundColor": {
                   "windowsbrush": [
-                    "SystemColorHighlightTextColor",
+                    "CardBackgroundFillColorDefaultBrush",
                   ],
                 },
               },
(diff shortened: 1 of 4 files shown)


__tests__/__snapshots__/App-test.js.snap
```diff
@@ -6171,7 +6191,7 @@ exports[`Expander Example Page 1`] = `
                       undefined,
                     ]
                   }
-                  timeZoneOffsetInSeconds={-0}
+                  timeZoneOffsetInSeconds={-28800}
```
marlenecota, line 6194: This seems odd

chiaramooney (commented):
Is it possible to use a color for the background that is different from the background of the app, so we can see the badge shape?

chrisglein:
> Is it possible to use a color for the background that is different from the background of the app, so we can see the badge shape?

Is possible. Figuring out the specific colors needs some looking through the design guides. I opened a separate issue for this since it's not regressed with this PR: https://github.com/microsoft/react-native-gallery/issues/421

chiaramooney (approved):

PR #413, opened 2024-04-19
Add "copy to clipboard" to code blocks

chrisglein:
### Why

WinUI3 gallery code blocks (and most code documentation website and tools) have a "copy to clipboard" button.

Previously:
image

### What

Adds a copy to clipboard button, and modifies the codeblock styling to match WinUI3 gallery. Clipboard was already a dependency that gallery had, so this is easy to add.

## Screenshots

image

image

### DIFF
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -252,7 +252,7 @@ exports[`Button Example Page 1`] = `
             style={
               {
                 "borderColor": "rgb(216, 216, 216)",
-                "borderRadius": 2,
+                "borderRadius": 8,
                 "borderWidth": 1,
               }
             }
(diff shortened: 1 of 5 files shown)


marlenecota (approved):

Yajur-Grover:
Was running some more FastPass checks and looks like there are a couple of accessibility issues with the 'copy to clipboard' icon as shown below: 
image

Doesn't look like a huge issue - main issue seems to be the Control Type and LocalizedControlType properties are set to Custom, and the Name property is not defined/has an invalid character. We can create a separate PR resolving this and merge it in.

chrisglein:
> Doesn't look like a huge issue - main issue seems to be the Control Type and LocalizedControlType properties are set to Custom, and the Name property is not defined/has an invalid character. We can create a separate PR resolving this and merge it in.

Thanks for checking. I used "Reference in a new issue" to create a linked issue to this and assigned to me.

Issue #405, opened 2024-04-10
Rejuvenate visuals to match current WinUI gallery


chrisglein:
When React Native Gallery was created, it was before the WinUI gallery had experienced a significant visual rejuvenation. We should be able to show the same class of visuals with a bit of effort. Here are the high impact items I can identify.

### Splash image and set of useful links
![image](https://github.com/microsoft/react-native-gallery/assets/26607885/acfbc8ae-5bf4-4baa-8a51-9cf43b93e996)
Many of these are just as relevant within RN Gallery
- Getting started (can be link to https://aka.ms/reactnative or specifically the getting started subpage)
- Windows design
- Github
- Template Studio (are there still RNW templates in there?)
- Code samples (link to our samples repo)
- Partner Center

### Recently added/updated
![image](https://github.com/microsoft/react-native-gallery/assets/26607885/ebd28cd5-c2cc-4b71-ae73-782d7846440f)
![image](https://github.com/microsoft/react-native-gallery/assets/26607885/002ef55e-4209-4096-ae20-c445cb60e880)

### Design guidance
![image](https://github.com/microsoft/react-native-gallery/assets/26607885/e4c0a8d5-bb37-4d47-9014-52d682066d01)

### Control group subpages (with full color icons in page view)
![image](https://github.com/microsoft/react-native-gallery/assets/26607885/3cc20a86-e81b-4030-ab0a-f809a93b5ee9)
![image](https://github.com/microsoft/react-native-gallery/assets/26607885/3cf1fc23-4fb9-4149-9d96-c9eba541d595)

Contrast with RN Gallery which has groups only on the main page, and the icons are monochromatic:
![image](https://github.com/microsoft/react-native-gallery/assets/26607885/3e7a56a8-851b-41b8-a8a4-f834a3948513)

Work in progress snack to get started:
https://snack.expo.dev/4NjQuP5ias-bJa04RLYIk?platform=web
https://snack.expo.dev/wvKq-kD7PqlROjiiiswqx?platform=web

```[tasklist]
## Tasks
- [x] Add splash image to home page
- [x] Add list container for useful links
- [x] Add link to Getting started (https://aka.ms/reactnative)
- [x] Windows design (https://learn.microsoft.com/en-us/windows/apps/design/_)
- [x] Github (https://github.com/microsoft/react-native-windows/)
- [ ] Template Studio (https://marketplace.visualstudio.com/items?itemName=TemplateStudio.TemplateStudioForUWP)
[...]

PR #414, opened 2024-04-19
Change home page icons from text icons to color icons

chrisglein:
### Why
Part of #405

Here's the home page of the current RN gallery:
image

Here's the home page of the current WinUI gallery:
image

### What

- Added images from WinUI3 gallery
- Adjusted layout to match WinUI3 gallery
- Adjusted categories to match WinUI3 gallery

## Screenshots

image

WinUI3 gallery accessibility tree
image

RN gallery accessibility tree
image

```[tasklist]
### Accessibility
- [x] The element's ControlType requires valid values for SizeOfSet and PositionInSet. (all list items)
- [x] The Name must not include the same text as the LocalizedControlType. (for "Image control" list item)
- [x] The Name property must not include the element's control type. (for "Image control" list item)
- [x] The Name property must not contain any characters in the private Unicode range. (Device info)
```


Yajur-Grover (commented):
Ran fastpass on the home page of the branch and there were some accessibility issues that came up here: 
image

Are you seeing the same locally?

chrisglein:
> Ran fastpass on the home page of the branch and there were some accessibility issues that came up here: 
> Are you seeing the same locally?

Hadn't run it yet but I'm seeing them now. I'll dig into these. I was trying to mirror the WinUI3 gallery accessibility, so I'll see how they handled these.

chrisglein:
> > Ran fastpass on the home page of the branch and there were some accessibility issues that came up here:
> > Are you seeing the same locally?
> 
> Hadn't run it yet but I'm seeing them now. I'll dig into these. I was trying to mirror the WinUI3 gallery accessibility, so I'll see how they handled these.
> 
> Edit: Added tasklist in issue description to track

Okay, pushed commits to address these and have it run clean.
- The "Image" name one triggers in WinUI3 Gallery too. I opted to workaround by using a different word ("Graphic"), which is not great but quiets the tool. I'll follow up to figure out if there's a way to suppress that warning on a specific item. Alternatively, we could live with the warning.
- I changed the layered image+text to show as an "image", because that's how the eye sees them.
- The last problem was that it didn't like the "list item" types because they were missing information about position/order. I reverted to "button" for now (what it was before). Will need some digging to figure out if list item is implemented fully or not.

chiaramooney (approved):
👍

PR #415, opened 2024-04-19
Add WinRT example page

Yajur-Grover:
### Why

### What

## Screenshots

### DIFF
--- a/src/RNGalleryList.ts
+++ b/src/RNGalleryList.ts
@@ -36,6 +36,7 @@ import {TrackPlayerExamplePage} from './examples/TrackPlayerExamplePage';
 import {WindowsHelloExamplePage} from './examples/WindowsHelloExamplePage';
 import {ExpanderExamplePage} from './examples/ExpanderExamplePage';
 import {VirtualizedListExamplePage} from './examples/VirtualizedListExamplePage';
+import {WinRTExamplePage} from './examples/WinRTExamplePage';
 
 let RNGalleryCategories = [
   'Basic Input',
@@ -264,6 +265,12 @@ export const RNGalleryList: Array<IRNGalleryExample> = [
     key: 'VirtualizedList',
     component: VirtualizedListExamplePage,
     icon: '\uE8A4',
+    type: 'Layout',
+  },
+  {
+    key: 'WinRT',
+    component: WinRTExamplePage,
+    icon: '\uEF15',
     type: 'Layout',
   },
   {
--- a/src/components/Notifications.tsx
+++ b/src/components/Notifications.tsx
@@ -1,0 +1,55 @@
+const Notifications = Windows.UI.Notifications;
+const ToastTemplateType = Notifications.ToastTemplateType;
+const ToastNotificationManager = Notifications.ToastNotificationManager;
+const ToastNotification = Notifications.ToastNotification;
+
+export function showNotification(notification) {
+  var type = ToastTemplateType.toastText01;
+
+  var obj = {};
+  if (typeof notification === 'string') {
+    obj.text = notification;
+  } else {
+    obj = notification;
+  }
+
+  if (obj.template != undefined) {
+    type = obj.template;
+  }
+
+  var xml = ToastNotificationManager.getTemplateContent(type);
+  for (var tagName in obj) {
+    var xmlElements = xml.getElementsByTagName(tagName);
+    var value = obj[tagName];
+    if (typeof value === 'string') {
+      fillXmlElements(xml, xmlElements, [value]);
+    } else if (Array.isArray(value)) {
+      fillXmlElements(xml, xmlElements, value);
+    } else if (typeof value === 'object') {
+      fillXmlElements(xml, xmlElements, [value]);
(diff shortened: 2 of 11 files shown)


src/RNGalleryList.ts
```diff
@@ -266,6 +267,12 @@ export const RNGalleryList: Array<IRNGalleryExample> = [
     icon: '\uE8A4',
     type: 'Layout',
   },
+  {
+    key: 'WinRT',
+    component: WinRTExamplePage,
+    icon: '\uEF15',
+    type: 'Layout',
+  },
```
chrisglein, line 275: FYI when you have to merge this later, look for:
`icon` -> `textIcon`

src/RNGalleryList.ts
```diff
@@ -266,6 +267,12 @@ export const RNGalleryList: Array<IRNGalleryExample> = [
     icon: '\uE8A4',
     type: 'Layout',
   },
+  {
+    key: 'WinRT',
+    component: WinRTExamplePage,
+    icon: '\uEF15',
+    type: 'Layout',
```
chrisglein, line 274:
```suggestion
    type: 'System',
```

src/components/Notifications.tsx
```diff
@@ -0,0 +1,55 @@
+const Notifications = Windows.UI.Notifications;
+const ToastTemplateType = Notifications.ToastTemplateType;
+const ToastNotificationManager = Notifications.ToastNotificationManager;
+const ToastNotification = Notifications.ToastNotification;
+
+export function showNotification(notification) {
+  var type = ToastTemplateType.toastText01;
```
chrisglein, line 7:
```suggestion
  let type = ToastTemplateType.toastText01;
```
I'm surprised we don't have lint rules setup about `var`. As a default in our JS, we should always prefer `let`. It's scoped in a way that is much more sensible.

src/components/Notifications.tsx
```diff
@@ -0,0 +1,55 @@
...
+  for (var tagName in obj) {
+    var xmlElements = xml.getElementsByTagName(tagName);
+    var value = obj[tagName];
+    if (typeof value === 'string') {
+      fillXmlElements(xml, xmlElements, [value]);
+    } else if (Array.isArray(value)) {
+      fillXmlElements(xml, xmlElements, value);
+    } else if (typeof value === 'object') {
+      fillXmlElements(xml, xmlElements, [value]);
+    }
```
chrisglein, line 30: Looks like case 1 and 3 are the same. Is this still correct?
```suggestion
    if (Array.isArray(value)) {
      fillXmlElements(xml, xmlElements, value);
    } else {
      fillXmlElements(xml, xmlElements, [value]);
    }
```
Or do you get types other than string/object in there? (like numeric literals or something)

chrisglein: If your goal is "if it's an array, leave it; otherwise, place in an array" there might be other options. Like:
```
[...value]
```
Does that do the right thing? It definitely spreads a source array into a new array. But for a single item... I think still the same thing? Hmmm... except strings are iterable so that'll break apart the string. Anyway, might be some syntactic magic here.

src/components/Notifications.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  var toast = new ToastNotification(xml);
+  ToastNotificationManager.createToastNotifier().show(toast);
+}
+
+function fillXmlElements(xml, xmlElements, arr) {
+  var i = 0;
+  for (var arrValue of arr) {
+    var node = xmlElements[i++];
```
chrisglein, line 40: What's happening here? Are `arr` and `xmlElements` ensured to be the same length? Why are we enumerating over `arr` but indexing into `xmlElements`?

src/examples/WinRTExamplePage.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return (
+    <Page
+      title="WinRT"
+      description="A React Native Windows module that allows use of native (non-XAML) WinRT APIs"
+      componentType="Community"
+      pageCodeUrl="https://github.com/microsoft/react-native-gallery/blob/main/src/examples/WinRTExamplePage.tsx"
+      documentation={[
+        {
+          label: 'WinRT Source Code',
+          url: 'https://github.com/asklar/react-native-winrt',
```
chrisglein, line 75:
```suggestion
          url: 'https://github.com/microsoft/react-native-winrt',
```

src/examples/WinRTExamplePage.tsx
```diff
@@ -0,0 +1,121 @@
...
+        },
+      ]}>
+      <Example title="A Windows Notification example" code={example1jsx}>
+        <View
+          style={{
+            flexDirection: 'row',
+            flexWrap: 'wrap',
+            alignItems: 'center',
+          }}>
+          <Pressable
```
chrisglein, line 85: Any reason this doesn't just use a `Button` instead of a custom Pressable? It looks... button-y. And the point of the sample isn't any exact visual style, so it should be simple first.

Issue #408, opened 2024-04-17
Phantom tab stops after content


will-ks:
If you press the tab key until focus is on the last element on the screen and then press tab again, tab focus should wrap around to the first element on screen. However, what actually happens is that the focus indicator disappears until you've pressed tab again about 30 times, then it finally wraps around to the start again. If you're running Narrator while you do this, the screen reader is completely silent on these phantom tab stops.

I'm posting this here because we are facing a similar problem in our app (although in ours, we only have a single phantom tab stop). I think it is to do with the windows implementation of react-navigation (in our app, removing the react-navigation `Stack.Navigator` makes the issue go away). Interested to see if anyone has any suggestions of what might be causing this, or any tips for debugging such an issue.

chrisglein:
@Yajur-Grover this sounds similar to some of the accessibility issues you've been investigating. Can you help provide guidance, share what you've learned (and/or the known bugs being worked on)?

Yajur-Grover:
I was able to repro the same behaviour, but it doesn't seem to be similar to any recent accessibility issue. This behavior is present both with or without Narrator being enabled. I also tried this on another example page in an app without react-navigation and there were no phantom stops. Will try and debug what the tab stop is being activated on when out of view.

Yajur-Grover:
Okay - it looks like the phantom tab stops are actually going through the navigation drawer. If you keep tabbing after the end of the page content and press 'Enter', depending on how many times you have pressed 'Tab', it will select the DrawerItem that is focused on currently, and open the component example page corresponding to the item.

chrisglein:
Made some edits to Gallery to clean up the drawer. After that, in a release build, there is still exactly one "phantom tab stab" (I like that term 👻). Needs digging in to find out what it is.

To the question posed above:
> Interested to see if anyone has any suggestions of what might be causing this, or any tips for debugging such an issue.

That is a key question, and worth spending some time documenting. In the past the way I've debugged this is with adding breakpoints/tracepoints to the various focus change events in the framework. Which is tedious at best. We should find out from the XAML team if there's helpful tooling for debugging "where has focus gone?", because it's not an uncommon problem.

will-ks:
[...]

Issue #407, opened 2024-04-16
Warning: All focusable views should report proper accessibility information. Views marked as focusable should always be accessible.


chrisglein:
App with no changes reports an accessibility warning.

![image](https://github.com/microsoft/react-native-gallery/assets/26607885/a6d8549a-c8dd-457f-94b0-7e1ef2744bfb)

```
 WARN  All focusable views should report proper accessibility information. Views marked as focusable should always be accessible.
    at View (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:59728:43)
    at Pressable (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:120292:27)
    at anonymous (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:75975:62)
    at PlatformPressable (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:183305:25)
    at LinkPressable (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:181675:21)
    at RCTView
    at View (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:59728:43)
    at DrawerItem (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:181725:93)
    at RCTScrollContentView
    at RCTScrollView
    at ScrollView (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:70087:36)
    at ScrollView
    at RCTView
    at View (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:59728:43)
    at CustomDrawerContent (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:125530:118)
    at RCTView
    at View (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:59728:43)
    at AnimatedComponent(View) (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:187128:38)
    at RCTView
    at View (http://localhost:8081/index.bundle//&platform=windows&dev=true&hot=true&inlineSourceMap=true:59728:43)
[...]

PR #417, opened 2024-04-22
Disable Gallery drawer content when closed

chrisglein:
### Why

When you tab through Gallery, you hit the closed drawer items. Those items should not be keyboard focusable.

Related to #408

### What

Removed the drawer content entirely when the drawer is in the closed state.
Additionally changed the drawer mode to not be "permanent", which didn't seem to be the cause here but also doesn't seem to be necessary.

## Screenshots

Here is gallery with the fix:
https://github.com/microsoft/react-native-gallery/assets/26607885/fcf25e92-450d-4627-8f4d-c8027e3a8f88

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -91,6 +91,10 @@ function RenderDrawer(props) {
 function CustomDrawerContent(props) {
   const isDrawerOpen =
     getDrawerStatusFromState(props.navigation.getState()) === 'open';
+
+  if (!isDrawerOpen) {
+    return <View />;
+  }
   return (
     <View style={styles.drawer}>
       <TouchableHighlight
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -164,9 +164,7 @@ function MyDrawer() {
   let screens = renderScreens();
   return (
     <Drawer.Navigator
-      drawerContent={(props) => (
-        <CustomDrawerContent {...props} drawerType="permanent" />
-      )}
+      drawerContent={(props) => <CustomDrawerContent {...props} />}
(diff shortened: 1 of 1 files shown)


Yajur-Grover:
Tested it locally - looks like there's still one phantom tab stop now on the home page: https://microsoft-my.sharepoint-df.com/:v:/p/yajurgrover/EY3otnRX6_tKu9cnSEYDMkkBdxn69UquorPhvm6M5LkdfA?e=yVEeMi&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0%3D

From the focusable warning call stack, looks like it's still focusing on a Drawer item?
image

chrisglein:
> Tested it locally - looks like there's still one phantom tab stop now on the home page: 
Yep, I'm still seeing one. So it matches #408 (instead of having a ton). Of note, you can get some if you have some logbox warnings. I'm going to check a release build and see if that removes the last tab stop. If not... well it's still cutting down on the many drawer ones.

> From the focusable warning call stack, looks like it's still focusing on a Drawer item? 
Yep, that's preexisting (#407). I'm working on a change to refactor the drawer entirely and I'm hoping we can fix that error at the same time.

chiaramooney (approved):

PR #424, opened 2024-04-24
Update native badge color

### LINKED ISSUE #421: "Wrapped Windows Control" badge does not show background/border in light mode
Is it possible to use a color for the background that is different from the background of the app, so we can see the badge shape?

_Originally posted by @chiaramooney in https://github.com/microsoft/react-native-gallery/pull/412#pullrequestreview-2018238452_

image

chrisglein:
### Why

Resolves #421

### What

Changes to a different brush that creates a bit more distinction between background color (but is ultimately still subtle).

## Screenshots

Chosen: `SolidBackgroundFillColorSecondaryBrush`
image
image

For reference (not chosen): `LayerFillColorAltBrush`
image
image

### DIFF
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -1679,7 +1679,7 @@ exports[`CheckBox Example Page 1`] = `
               {
                 "backgroundColor": {
                   "windowsbrush": [
-                    "LayerFillColorAltBrush",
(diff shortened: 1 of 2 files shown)


Yajur-Grover (approved):
LGTM - looks like the color contrast is above the minimum required!

marlenecota (approved):

PR #439, opened 2024-05-01
chore: use Yarn 4

Saadnajmi:
### Why

Yarn 4 exists and is better and compatible, why not use it? :)

### What

Ran the command `yarn set version berry`, and updated the gitignore to ignore `install-state.gz`. Interestingly, yarn couldn't find `"@types/": "react-navigation/drawer",` so it just removed that from our package.json

## Screenshots

N/A


Saadnajmi:
Ironically, this PR makes adding macOS support harder, because `react-native-macos-init` doesn't support yarn 2+ yet. Good forcing function :D

chiaramooney (approved):

chiaramooney:
@Saadnajmi We can merge as soon as the merge conflicts are resolved.

PR #470, opened 2024-07-09
Refactor: Fetch high contrast state via a cross platform hook

Saadnajmi:
### Why

This is part of a series of commits to reduce dependence on React Native Windows specific APIs, and allow the gallery to potentially run on macOS. 

Currently, `App.tsx` uses a React Native Windows specific JS API to fetch the high contrast state. This adds a direct import to `react-native-windows` that isn't gated inside a platform check or file. This commit refactors the code to fetch high contrast state into a hook, so that the windows specific imports can be moved to a `foo.windows.tsx` file, and so that we can also implement the macOS / mobile implementations separately. 

### What

Moves high contrast fetching code to a new hook `useHighContrastState()`

## Screenshots

No change should be seen to final app. 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/470)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,8 +22,8 @@ import {
   ThemeSetterContext,
 } from './themes/Theme';
 import {PlatformColor} from 'react-native';
-import {AppTheme} from 'react-native-windows';
 import HighContrastTheme from './themes/HighContrastTheme';
+import useHighContrastState from './hooks/useHighContrastState';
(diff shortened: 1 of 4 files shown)


chiaramooney (approved):
👍

PR #471, opened 2024-07-09
Fix: Use `@rnx-kit/metro-config`

Saadnajmi:
### Why

The current metro config contains some RNW specifics, and doesn't work with React Native macOS. 

### What

Replace the metro config with the one provided by rnx-kit, which is more cross platform friendly.

## Screenshots

No change to final app. 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/471)


chiaramooney (approved):
👍

chiaramooney:
Making a note here that since we are editing the metro config to remove some of the Windows-isms there might be some edge case scenarios that will break. I'm not sure how much we have tested the rnx-kit/metro-config. If we start seeing metro issues, we should check if this change is the cause.

PR #472, opened 2024-07-09
Fix: Update `react-native-xaml` to remove `@types/react-native` dependency

Saadnajmi:
### Why

React Native ships its' own typescript types now, and the ones in the old package `@types/react-native` may conflict. We are bringing in the old package because our version of `react-native-xaml` depends on it. However, as of https://github.com/microsoft/react-native-xaml/commit/d9e933a30b836511609fd5a09acdafd3b9c20007 the dependency is removed. 

### What

Update our `react-native-xaml` dependency and remove the old types package from our lock. 

## Screenshots

Should be no change to final app. 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/472)

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "react-native-tts": "ak1394/react-native-tts",
     "react-native-webview": "^13.2.2",
     "react-native-windows": "0.74.2",
-    "react-native-xaml": "^0.0.74"
+    "react-native-xaml": "^0.0.78"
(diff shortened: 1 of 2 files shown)


chiaramooney (approved):
👍

PR #475, opened 2024-07-16
Add/fix `accessibilityLabel` property values for Clipboard, XAML and Lottie Example pages

### LINKED ISSUE #453: 'Name of a focuable element must not be null' error message from Accessibility Insights on Clipboard example page
Screenshot of Fastpass:

image

Looks like we need to have the accessibilityLabel prop defined on each of these.

### LINKED ISSUE #454: 'Name of a focuable element must not be null' error message from Accessibility Insights on Lottie example page
Screenshot of Fastpass: 

image

Looks like we need to have the accessibilityLabel prop defined on each of these.

### LINKED ISSUE #474: UIA Name and ControlType/LocalizedControlType property errors on Xaml example change.
Screenshot of Fastpass:

image

All errors are due to the UIA Name and UIA ControlType and LocalizedControlType properties having common text in their values. Will need to change the UIA Name properties (accessibilityLabel) for the impacted components.

Yajur-Grover:
## Description
Fix accessibilityLabel values to prevent issues with UIA ControlType/LocalizedControlType values being the same, or UIA Name property values being null on Lottie, Clipboard and XAML example pages.

### Why

Resolves #453 
Resolves #454 
Resolves #474 

### What

## Screenshots
Clipboard before:
image

Clipboard After:
Clipboard After

Lottie Before:
image

Lottie After: 
Lottie After

XAML Before:
image

XAML After:
XAML After

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/475)

### DIFF
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -3719,6 +3719,7 @@ exports[`Clipboard Example Page 1`] = `
                 }
               >
                 <View
+                  accessibilityLabel="Example copy text"
                   accessibilityRole="button"
                   accessibilityState={
                     {
@@ -3801,6 +3802,7 @@ exports[`Clipboard Example Page 1`] = `
                   </View>
                 </View>
                 <TextInput
+                  accessibilityLabel="Example set text to copy"
                   onChangeText={[Function]}
                   value="This text will be copied to the clipboard"
                 />
@@ -4002,6 +4004,7 @@ exports[`Clipboard Example Page 1`] = `
                 }
               >
                 <View
+                  accessibilityLabel="Example paste text"
                   accessibilityRole="button"
                   accessibilityState={
                     {
@@ -54836,6 +54839,7 @@ exports[`Xaml Example Page 1`] = `
               }
             >
               <XamlControl
+                accessibilityLabel="Simple example"
                 content={
                   {
                     "string": "XAML Button",
@@ -56349,7 +56353,7 @@ You selected Option 1
               }
             >
               <XamlControl
-                accessibilityLabel="Simple example ComboBox"
+                accessibilityLabel="Simple example"
                 text="ComboBox"
                 type="Windows.UI.Xaml.Controls.ComboBox"
               >
--- a/src/examples/ClipboardExamplePage.tsx
+++ b/src/examples/ClipboardExamplePage.tsx
@@ -43,16 +43,22 @@ export const ClipboardExamplePage: React.FunctionComponent<{}> = () => {
       <Example title="Copy text to the Clipboard." code={example1jsx}>
         <View style={{alignItems: 'flex-start', gap: 12}}>
           <Button
+            accessibilityLabel="Example copy text"
             title="Copy text to the Clipboard"
             onPress={() => Clipboard.setString(textToCopy)}
           />
-          <TextInput value={textToCopy} onChangeText={setTextToCopy} />
+          <TextInput
+            accessibilityLabel="Example set text to copy"
+            value={textToCopy}
+            onChangeText={setTextToCopy}
+          />
         </View>
       </Example>
 
       <Example title="Paste text from the Clipboard." code={example2jsx}>
         <View style={{alignItems: 'flex-start', gap: 12}}>
           <Button
+            accessibilityLabel="Example paste text"
             title="Paste text from the Clipboard"
             onPress={() => getClipboardText()}
           />
--- a/src/examples/LottieAnimationsExamplePage.tsx
+++ b/src/examples/LottieAnimationsExamplePage.tsx
@@ -100,15 +100,11 @@ export const LottieAnimationsExamplePage: React.FunctionComponent<{}> = () => {
           }}
         />
         <View style={{flexDirection: 'row', gap: 12}}>
-          <Button onPress={onPlay} title={'play'} accessibilityLabel="Play" />
-          <Button
-            onPress={onPause}
-            title={'pause'}
-            accessibilityLabel="Pause"
-          />
+          <Button onPress={onPlay} title="play" accessibilityLabel="Play" />
+          <Button onPress={onPause} title="pause" accessibilityLabel="Pause" />
           <Button
             onPress={onResume}
-            title={'resume'}
+            title="resume"
             accessibilityLabel="Resume"
           />
           <Button onPress={onReset} title="reset" accessibilityLabel="Reset" />
--- a/src/examples/LottieAnimationsExamplePage.tsx
+++ b/src/examples/LottieAnimationsExamplePage.tsx
@@ -100,13 +100,22 @@ export const LottieAnimationsExamplePage: React.FunctionComponent<{}> = () => {
           }}
         />
         <View style={{flexDirection: 'row', gap: 12}}>
-          <Button onPress={onPlay} title={'play'} />
-          <Button onPress={onPause} title={'pause'} />
-          <Button onPress={onResume} title={'resume'} />
-          <Button onPress={onReset} title="reset" />
+          <Button onPress={onPlay} title={'play'} accessibilityLabel="Play" />
+          <Button
+            onPress={onPause}
+            title={'pause'}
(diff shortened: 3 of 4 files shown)


__tests__/__snapshots__/App-test.js.snap
```diff
@@ -54836,6 +54839,7 @@ exports[`Xaml Example Page 1`] = `
               }
             >
               <XamlControl
+                accessibilityLabel="Simple example"
```
marlenecota, line 54842:
```suggestion
                accessibilityLabel="Simple button example"
```

__tests__/__snapshots__/App-test.js.snap
```diff
@@ -56349,7 +56353,7 @@ You selected Option 1
               }
             >
               <XamlControl
-                accessibilityLabel="Simple example ComboBox"
+                accessibilityLabel="Simple example"
```
marlenecota, line 56356:
```suggestion
                accessibilityLabel="Simple combobox example"
```

src/examples/LottieAnimationsExamplePage.tsx
```diff
@@ -100,13 +100,22 @@ export const LottieAnimationsExamplePage: React.FunctionComponent<{}> = () => {
...
+          />
+          <Button
+            onPress={onResume}
+            title={'resume'}
+            accessibilityLabel="Resume"
+          />
+          <Button onPress={onReset} title="reset" accessibilityLabel="Reset" />
           <Button
             onPress={handlerChangeLoop}
             title={loop ? 'disable loop' : 'ative resume'}
```
marlenecota, line 117: "ative"? is this a typo?

src/examples/LottieAnimationsExamplePage.tsx
```diff
@@ -100,13 +100,22 @@ export const LottieAnimationsExamplePage: React.FunctionComponent<{}> = () => {
...
+            accessibilityLabel="Disable Loop"
```
marlenecota, line 118: 'ative loop' can't be right, so I just put 'resume loop', but whatever it should be, it should match the title. 

```suggestion
            accessibilityLabel={loop ? 'disable loop' : 'resume loop'}
```

src/examples/XamlExamplePage.tsx
```diff
@@ -113,7 +113,11 @@ const onRadioButtonSelected = (event) => {
         <TextBlock text="I am a XAML TextBlock." />
       </Example>
       <Example title="A simple XAML Button." code={example2jsx}>
-        <Button content={{string: 'XAML Button'}} onClick={() => {}} />
+        <Button
+          accessibilityLabel="Simple example"
```
marlenecota, line 117:
```suggestion
          accessibilityLabel="Simple button example"
```

src/examples/XamlExamplePage.tsx
```diff
@@ -152,7 +156,7 @@ You selected Option ${selectedRadioButton}`}</Text>
         </View>
       </Example>
       <Example title="A simple ComboBox." code={example5jsx}>
-        <ComboBox accessibilityLabel="Simple example ComboBox" text="ComboBox">
+        <ComboBox accessibilityLabel="Simple example" text="ComboBox">
```
marlenecota, line 159:
```suggestion
        <ComboBox accessibilityLabel="Simple combobox example" text="ComboBox">
```

Yajur-Grover:
@marlenecota for the accessibilityLabel changes, removing the component name for the label is intentional, since that is what is causing the issue within the UIA tree. Since the screen reader reads out the 'Name' property and 'ControlType' property consecutively, the 'Name' property cannot have the same value as the ControlType, since the screen reader will read out the same value twice. 

E.g if `accessibilityLabel='Simple combobox example'`, then `Name='Simple combobox example'` and `ControlType='Combobox'`, and the screenreader will say `'Simple combobox example, combobox'` which is a problem for understandability.

marlenecota (approved):

src/examples/LottieAnimationsExamplePage.tsx
Yajur-Grover: Most likely - applied the recommended change.

src/examples/ClipboardExamplePage.tsx
```diff
@@ -43,16 +43,22 @@ export const ClipboardExamplePage: React.FunctionComponent<{}> = () => {
       <Example title="Copy text to the Clipboard." code={example1jsx}>
         <View style={{alignItems: 'flex-start', gap: 12}}>
           <Button
+            accessibilityLabel="Example copy text"
```
chrisglein, line 46: Why doesn't Button.title become the default accessibilityLabel?

src/examples/LottieAnimationsExamplePage.tsx
```diff
@@ -100,13 +100,22 @@ export const LottieAnimationsExamplePage: React.FunctionComponent<{}> = () => {
...
-          <Button onPress={onReset} title="reset" />
+          <Button onPress={onPlay} title={'play'} accessibilityLabel="Play" />
+          <Button
+            onPress={onPause}
+            title={'pause'}
+            accessibilityLabel="Pause"
+          />
+          <Button
+            onPress={onResume}
+            title={'resume'}
```
chrisglein, line 111:
```suggestion
            title="resume"
```

src/examples/LottieAnimationsExamplePage.tsx
```diff
@@ -100,13 +100,22 @@ export const LottieAnimationsExamplePage: React.FunctionComponent<{}> = () => {
...
         />
         <View style={{flexDirection: 'row', gap: 12}}>
-          <Button onPress={onPlay} title={'play'} />
-          <Button onPress={onPause} title={'pause'} />
-          <Button onPress={onResume} title={'resume'} />
-          <Button onPress={onReset} title="reset" />
+          <Button onPress={onPlay} title={'play'} accessibilityLabel="Play" />
+          <Button
+            onPress={onPause}
+            title={'pause'}
```
chrisglein, line 106:
```suggestion
            title="pause"
```

src/examples/LottieAnimationsExamplePage.tsx
```diff
@@ -100,13 +100,22 @@ export const LottieAnimationsExamplePage: React.FunctionComponent<{}> = () => {
           }}
         />
         <View style={{flexDirection: 'row', gap: 12}}>
-          <Button onPress={onPlay} title={'play'} />
-          <Button onPress={onPause} title={'pause'} />
-          <Button onPress={onResume} title={'resume'} />
-          <Button onPress={onReset} title="reset" />
+          <Button onPress={onPlay} title={'play'} accessibilityLabel="Play" />
```
chrisglein, line 103:
```suggestion
          <Button onPress={onPlay} title="play" accessibilityLabel="Play" />
```

src/examples/ClipboardExamplePage.tsx
Yajur-Grover: I think it's because of how we define the accessibilityLabel property [here](https://github.com/microsoft/react-native-windows/blob/63d919e36e0d6ad93f1e772df53a1e7c1a5aa1da/vnext/src-win/Libraries/Components/Button.windows.js#L378). 

If we set the `accessibilityLabel` property based on the `title` instead would probably automatically do that. If the `title` property contains text that is the same as the ControlType (which would be 'button' here), then it's up to the client to add an appropriate `accessibilityLabel` property. Can open up a PR with this change if this approach makes sense.

chrisglein: It seems to match the cross-platform implementation:
https://github.com/facebook/react-native/blob/a6f5e5adebed3d9da411f99548e2d8ce96636e16/packages/react-native/Libraries/Components/Button.js#L366

It seems like a bad default to me personally... but we're matching core and that's what we should do. As in this is the same JS you'd use on mobile. It's just... duplicative and verbose. Oh well.

chrisglein (approved):

PR #476, opened 2024-07-16
Remove unused `doNotInset` prop

Saadnajmi:
### Why

This prop is always true. Let's just remove it. 

### What

Remove the prop `doNotInset` from ScreenWrapper

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/476)

### DIFF
--- a/src/ComponentListPage.tsx
+++ b/src/ComponentListPage.tsx
@@ -106,7 +106,7 @@ const ComponentListPage = ({route, navigation}: ComponentListPageProps) => {
 
   return isScreenFocused ? (
     <View>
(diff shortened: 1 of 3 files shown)


chrisglein:
> This prop is always true. Let's just remove it.

Not quite. It's left at the default for the default `Page` and `SettingsPage` in particular. I added this bool to separate full-bleed pages (that needed to inset their content) and just-content pages (which should be offset from the left bar).

Saadnajmi:
Oh, thanks! I'll close.

PR #489, opened 2024-09-09
Fix tile gallery link accessibility

### LINKED ISSUE #446: Accessibility issue with home page's link tiles
image

image

# Section 508 502.3.1: (rngallery/link '  Getting startedAn overview of app development options, tools, and samples.') The Name property must not contain any characters in the private Unicode range.

**App:** rngallery

**Element path:** link '  Getting startedAn overview of app development options, tools, and samples.'

**Issue Details:** The Name property must not contain any characters in the private Unicode range. [Section 508 502.3.1](https://www.access-board.gov/ict/#502-interoperability-assistive-technology)

**How To Fix:** Modify the Name property, removing all characters in the range U E000 to U F8FF and replace them with meaningful, standard text content.

# Section 508 502.3.10: (rngallery/link '  Getting startedAn overview of app development options, tools, and samples.') An element of the given ControlType must support the Invoke pattern.

**App:** rngallery

**Element path:** link '  Getting startedAn overview of app development options, tools, and samples.'

**Issue Details:** An element of the given ControlType must support the Invoke pattern. [Section 508 502.3.10](https://www.access-board.gov/ict/#502-interoperability-assistive-technology)

**How To Fix:** 1. Make sure the element has the appropriate ControlType property for its function.
2. If the current ControlType is correct, modify the element to support the Invoke pattern.

This accessibility issue was found using Accessibility Insights for Windows, a tool that helps debug and find accessibility issues earlier. Get more information and download this tool at https://aka.ms/AccessibilityInsights.

chrisglein:
## Description

Fix some issues with the tile gallery links

### Why

Resolves #446

### What

- Manually specify an accessibility label, avoiding the auto concatenation of all elements that introduced the link character. 
- Specifying a "link" role gives the right type in the inspector... but doesn't come with an invoke action. Falling back to "button" no longer matches the WinUI 3 Gallery, but passes the accessibility inspection. Workaround for https://github.com/microsoft/react-native-windows/issues/13861

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/489)

### DIFF
--- a/src/components/TileGallery.tsx
+++ b/src/components/TileGallery.tsx
@@ -80,7 +80,8 @@ const HeaderTile = (props: HeaderTileType): JSX.Element => {
       onPressOut={() => setIsPressing(false)}
       onHoverIn={() => setIsHovered(true)}
       onHoverOut={() => setIsHovered(false)}
-      accessibilityRole="link">
+      accessibilityRole="link"
+      accessibilityLabel={props.title}>
       <View style={styles.tileIconContent}>{props.children}</View>
       <Text style={styles.tileTitle}>{props.title}</Text>
       <Text style={styles.tileDescription}>{props.description}</Text>
--- a/src/components/TileGallery.tsx
+++ b/src/components/TileGallery.tsx
@@ -80,7 +80,7 @@ const HeaderTile = (props: HeaderTileType): JSX.Element => {
       onPressOut={() => setIsPressing(false)}
       onHoverIn={() => setIsHovered(true)}
       onHoverOut={() => setIsHovered(false)}
-      accessibilityRole="link"
+      accessibilityRole="button"
       accessibilityLabel={props.title}>
       <View style={styles.tileIconContent}>{props.children}</View>
       <Text style={styles.tileTitle}>{props.title}</Text>


chrisglein:
@chiaramooney is there a way to get the InvokeAction to get set up while having the role type set to "link"? It works by default for "button" but not "link".
(this is on Paper, but I assume that with implementation in place the same question would stand for Fabric)

chiaramooney:
So specifying the "link" accessibilityRole type should map to the UIA control type Hyperlink which according to this source (https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-controlpatternmapping) should always support the invoke provider. I would confirm using Accessibility Insights if the correct UIA control type is being specified. If not, then our role to control type mapping on Paper has a bug. If the control type is Hyperlink but the invoke provider is not implemented then this is a bug on Paper with our UIA Provider code (something in DynamicAutomationProvider is wrong, potentially hyperlink is missing from the list of controls that should support invoke). 

Since we are redoing our accessibility implementation from scratch on Fabric this issue should not be occurring on the new arch.

chrisglein:
> So specifying the "link" accessibilityRole type should map to the UIA control type Hyperlink which according to this source (https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-controlpatternmapping) should always support the invoke provider. I would confirm using Accessibility Insights if the correct UIA control type is being specified. If not, then our role to control type mapping on Paper has a bug. If the control type is Hyperlink but the invoke provider is not implemented then this is a bug on Paper with our UIA Provider code (something in DynamicAutomationProvider is wrong, potentially hyperlink is missing from the list of controls that should support invoke).
> 
> Since we are redoing our accessibility implementation from scratch on Fabric this issue should not be occurring on the new arch. See this line (https://github.com/microsoft/react-native-windows/blob/2c70908a390daa63060059fea373b4777f7d7b50/vnext/Microsoft.ReactNative/Fabric/Composition/CompositionDynamicAutomationProvider.cpp#L154C79-L154C106) where were create the IInvokeProvider if the role type is link.

Thanks Chiara, that's super helpful. I opened https://github.com/microsoft/react-native-windows/issues/13861.

Saadnajmi (approved):

Yajur-Grover (approved):
Looks good to me! Will try to take a stab at adding support for invoke to 'link'.

PR #492, opened 2024-09-26
Make page background and text visible in Desert and Aquatic high contrast themes

### LINKED ISSUE #491: Bug 54110592: [React Native Gallery]: All the content i.e., headers and static text is not visible in high contrast aquatic and desert themes.
**Build Details:**
App: React Native Gallery 
App Version: 1.0.18.0
AT: High Contrast 
 

**Pre-Requisite:**  

Turn on High Contrast Aquatic Theme (Settings->Accessibility->Contrast themes->Aquatic/Dessert theme->Apply)

**Repro-Steps:**
Navigate to hamburger button and expand it.
Now navigate to any button and select it.
Now observe the issue in high contrast themes.
Actual Result:
All the group's headers and some text present in the whole application is not visible in high contrast aquatic and desert theme.

**Expected Result:**
All the content i.e., headers and static text should be visible clearly in high contrast aquatic and desert theme.

**User impact:**

Users with low vision will not be able to consume the text if it is not visible in High Contrast Themes.

Link to ADO bug: https://microsoft.visualstudio.com/OS/_workitems/edit/54110592

Yajur-Grover:
## Description
Make background and text in example page headers and body visible in Desert and Aquatic high contrast themes.

### Why

Resolves #491 

### What

Change PlatformColor values for ScreenWrapper background color.

## Screenshots

Button example page in Aquatic theme:
image

Button example page in Desert theme:
image

### DIFF
--- a/src/components/Code.tsx
+++ b/src/components/Code.tsx
@@ -20,7 +20,7 @@ const darkColors = {
 
 const lightColors = {
   background: 'transparent',
-  text: '#000000',
+  text: '#FFFFFFF',
   comment: '#000000',
   keyword: '#FF0000',
   tag: '#A31515',
--- a/src/components/Code.tsx
+++ b/src/components/Code.tsx
@@ -1,12 +1,12 @@
 /* eslint-disable react-hooks/rules-of-hooks */
 import low from 'lowlight';
 import React from 'react';
-import {StyleSheet, Text, View} from 'react-native';
+import {PlatformColor, StyleSheet, Text, View} from 'react-native';
 import {ThemeContext} from '../themes/Theme';
 
 const darkColors = {
   background: 'transparent',
-  text: '#9cdcfe',
+  text: PlatformColor('SystemControlPageTextBaseMediumBrush'),
   comment: '#6a9955',
   keyword: '#c586c0',
   tag: '#cc99cc',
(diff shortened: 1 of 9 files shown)


src/components/Code.tsx
```diff
@@ -20,7 +20,7 @@ const darkColors = {
 
 const lightColors = {
   background: 'transparent',
-  text: '#000000',
+  text: '#FFFFFFF',
```
chrisglein, line 23: Why would the default text color for "light" theme be white?

Yajur-Grover: The text color of the example code blocks looks the same: 
image
Checked some other pages and the text color is the same.

Yajur-Grover: Changing the color fixed the issue in contrast mode - originally the text in the code block looked like this: 
image

chrisglein: But... it still doesn't make sense to me?
Also why would the value be changed for the snapshots that aren't in contrast mode? Shouldn't those stay on an unchanged color since the snapshots were of light/dark that worked fine before?

chrisglein: Some other potential lenses on this:
- Should these actually be `PlatformColor` values instead of these fixed colors?
- The logic below that selects colors based on theme only uses light/dark as a binary choice. How should that change to accommodate high contrast and high contrast variants?

Yajur-Grover: @chrisglein have changed the text color values to accomodate `PlatformColor`. By using a `SolidColorBrush` value, the color of the text should always accomodate any high contrast theme setting, per the documentation [here](https://microsoft.github.io/react-native-windows/docs/windowsbrush-and-theme#applying-a-system-accent-color-variant). I have tested this on the Desert and Aquatic theme modes, as well as in normal, and there are no visibility issues.

chrisglein (approved):

PR #521, opened 2024-10-24
Fix text on Home Page tiles being truncated too much when changing text scaling value

### LINKED ISSUE #503: Bug 54111342: [React Native Gallery -> Home]: The text present inside 'Get started/React Native/Windows Design/GitHub...' etc links are getting truncated after applying text scaling to 200%.
Build Details:
App: React Native Gallery 
App Version: 1.0.18.0

AT: Text Scaling  
Resolution: 1280*768

Pre-Requisite:
Text Scaling (Settings->Accessibility->Text size-> Drag the slider to 200% ->Apply))

Repro-Steps:
Navigate to ‘Home' button present in left navigation pane and invoke it.
Now observe the text present inside the 'Get started/React Native/Windows Design/GitHub...' etc links after applying text scaling.
Actual Result:
The text present inside 'Get started/React Native/Windows Design/GitHub...' etc links are getting truncated after applying text scaling to 200%.

Note: Same issue is observed in the entire application.

Expected Result:
Text present inside 'Get started/React Native/Windows Design/GitHub...' etc links should be visible clearly to the user without any data loss and truncation.

User impact:

Users with low vision will not be able to consume the content of the button if it is not visible in Text scaling mode.

Link to ADO bug: https://dev.azure.com/microsoft/OS/_workitems/edit/54111342

Yajur-Grover:
## Description
Adjusts `height` properties of affected Pressable components on home page to prevent text from being truncated too much when the text scaling value is changed and keep it visible to users.
### Why

Resolves #503 

### What

Adjust `{height}` properties of affected Pressable components.

## Screenshots

Before: 
image
After: 
image
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/521)

### DIFF
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -32394,6 +32394,7 @@ exports[`ProgressView Example Page 1`] = `
               }
             >
               <RNCProgressViewWindows
+                accessibilityLabel="Simple ProgressView"
                 progress={0.5}
                 style={
                   [
@@ -32585,6 +32586,7 @@ exports[`ProgressView Example Page 1`] = `
               }
             >
               <RNCProgressViewWindows
+                accessibilityLabel="Indeterminate ProgressView"
                 isIndeterminate="true"
                 style={
                   [
@@ -32776,6 +32778,7 @@ exports[`ProgressView Example Page 1`] = `
               }
             >
               <RNCProgressViewWindows
+                accessibilityLabel="Colored ProgressView"
                 progress={0.7}
                 progressTintColor="rgb(0, 122, 255)"
                 style={
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -3719,7 +3719,7 @@ exports[`Clipboard Example Page 1`] = `
                 }
               >
                 <View
-                  accessibilityLabel="Example copy text"
+                  accessibilityLabel="Copy text to the Clipboard"
                   accessibilityRole="button"
                   accessibilityState={
                     {
@@ -4004,7 +4004,7 @@ exports[`Clipboard Example Page 1`] = `
                 }
               >
                 <View
-                  accessibilityLabel="Example paste text"
+                  accessibilityLabel="Paste text from the Clipboard"
                   accessibilityRole="button"
                   accessibilityState={
                     {
--- a/__tests__/__snapshots__/App-test.js.snap
+++ b/__tests__/__snapshots__/App-test.js.snap
@@ -4,7 +4,11 @@ exports[`Button Example Page 1`] = `
 <View
   style={
     {
-      "backgroundColor": "#f9f9f9",
+      "backgroundColor": {
+        "windowsbrush": [
+          "SystemControlBackgroundListLowBrush",
+        ],
+      },
       "flexDirection": "row",
       "height": "100%",
       "width": "100%",
@@ -391,7 +395,11 @@ exports[`Button Example Page 1`] = `
                   selectable={true}
                   style={
                     {
-                      "color": "#000000",
+                      "color": {
+                        "windowsbrush": [
+                          "SystemControlPageTextBaseMediumBrush",
+                        ],
+                      },
                       "fontFamily": "Consolas",
                     }
                   }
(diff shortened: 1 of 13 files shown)


Yajur-Grover:
@chrisglein @chiaramooney Seems like a pretty hacky fix here to disable font scaling so that it remains visible, but the other two properties that I tried to use didn't seem to do anything. One alternative is to maybe try and get the entire parent component to scale with the text so that it fits - looked into initially and it doesn't seem like there's a directly accessible property that will allow this, but will try and see if there's some ways to play with style properties to achieve this.

chrisglein:
> @chrisglein @chiaramooney Seems like a pretty hacky fix here to disable font scaling so that it remains visible, but the other two properties that I tried to use didn't seem to do anything. One alternative is to maybe try and get the entire parent component to scale with the text so that it fits - looked into initially and it doesn't seem like there's a directly accessible property that will allow this, but will try and see if there's some ways to play with style properties to achieve this.

Disabling font scaling definitely isn't ideal. I think the root problem here is that the items are fixed sized containers that aren't prepared for the text to scale larger. Is it possible to get the text scale and then proportionally scale up the size of those containers?

Places like these: 
https://github.com/microsoft/react-native-gallery/blob/3d016aef467b587a7bd0f21d3fa346d374c26cfc/src/components/TileGallery.tsx#L33
https://github.com/microsoft/react-native-gallery/blob/3d016aef467b587a7bd0f21d3fa346d374c26cfc/src/components/ControlItem.tsx#L31

Yajur-Grover:
> Disabling font scaling definitely isn't ideal. I think the root problem here is that the items are fixed sized containers that aren't prepared for the text to scale larger. Is it possible to get the text scale and then proportionally scale up the size of those containers?
> 
 Changing the height to a percentage value works on the scaling issue for all the tiles on the home page except the 'Xaml' example - not sure why it is being set to a different value as it looks like it is generated using the same `HomeComponentTile` [component](https://github.com/microsoft/react-native-gallery/blob/3d016aef467b587a7bd0f21d3fa346d374c26cfc/src/ComponentListPage.tsx#L64) as part of the `<ListOfComponents>` on the [home page](https://github.com/microsoft/react-native-gallery/blob/3d016aef467b587a7bd0f21d3fa346d374c26cfc/src/HomePage.tsx#L128):

image

The Xaml tile is also smaller in size when font is set to a normal size.

chrisglein:
> image
> 
> The Xaml tile is also smaller in size when font is set to a normal size.

Overall that looks closer to what I'd expect for the behavior with larger font size. Maybe apply a [`minHeight`](https://reactnative.dev/docs/layout-props#minheight) instead of none.

No clue what's happening on the Xaml one. End of the scroller layout issue?

Yajur-Grover:
After adding minHeight values:

Tiles at 100%:
image

Tiles at 225%:
image

chrisglein:
> Tiles at 100% text scaling: image

The "ControlItem" buttons in the recently added lists look taller than they were before. Compare with:
image
Is the min height too tall?

Yajur-Grover:
I set the `minHeight` to 110, whereas previously the `height` variable was set to 90. With the height at 90, the XAML tile didn't appear to have enough padding as shown here: 
image

src/components/ControlItem.tsx
```diff
@@ -28,7 +28,8 @@ const createStyles = (colors: any, isHovered: boolean, isPressing: boolean) =>
       flexDirection: 'row',
       gap: 16,
       width: 360,
-      height: 90,
+      height: '50%',
+      minHeight: 110,
```
chrisglein, line 32:
```suggestion
      minHeight: 90,
```
I don't think you need the 50% height.

src/components/TileGallery.tsx
```diff
@@ -30,7 +30,8 @@ const createStyles = (isHovered: boolean, _isPressing: boolean) =>
       padding: 24,
       gap: 4,
       width: 198,
-      height: 220,
+      height: '100%',
+      minHeight: 220,
```
chrisglein, line 34:
```suggestion
      minHeight: 220,
```
I don't think you need the specified 100% height.

__tests__/__snapshots__/App-test.js.snap
```diff
@@ -32394,6 +32394,7 @@ exports[`ProgressView Example Page 1`] = `
               }
             >
               <RNCProgressViewWindows
+                accessibilityLabel="Simple ProgressView"
```
chrisglein, line 32397: FYI the snapshot changes seem unrelated to this PR. Missed from an earlier change?

chrisglein (approved):

PR #545, opened 2025-04-08
[BugFix] Add accessibilityLabel for icon

anupriya13:
### Why

Build Details:
App: React Native Gallery
App Version: 1.0.18.0
AT: Fast Pass

Repro-Steps:
Navigate to ‘Home’ button present in left navigation pane and invoke it.
Now run fast pass and observe the issue.
Actual Result:
The Name property is containing characters in the private Unicode range.

Expected Result:
The Name property must not contain any characters in the private Unicode range.

User Impact:
Users with visual impairment who rely on screen reader will be misguided if narrator announces incorrect control type of the item.

Link to ADO bug: https://microsoft.visualstudio.com/OS/_workitems/edit/54111816

Resolves https://github.com/microsoft/react-native-gallery/issues/518

### What

This should be able to be resolved by changing the value of the accessibilityLabel property of the affected component 

## Screenshots

image

### DIFF
--- a/NewArch/src/components/ScreenWrapper.tsx
+++ b/NewArch/src/components/ScreenWrapper.tsx
@@ -87,7 +87,7 @@ export function ScreenWrapper({
             onPress={() => navigation.dispatch(DrawerActions.openDrawer())}
             activeOpacity={0.5783}
             underlayColor="rgba(0, 0, 0, 0.0241);">
-            <Text style={styles.icon}>&#xE700;</Text>
+            <Text style={styles.icon} accessibilityLabel="Navigation bar hamburger icon text">&#xE700;</Text>
           </TouchableHighlight>
         </View>
       </Pressable>


Yajur-Grover:
Also - New Arch Gallery is now in main - you can merge changes for it into the main branch instead of the NewArch branch.

anupriya13:
> Do you have a screenshot of Accessibility Insights showing the error showing up before the change, and then not showing up after? Looks like the screenshot you shared is the after - can you add the before one in as well?

Yeah it's of after, previously the error came up and now didn't.. will update before screenshot tomorrow

Yajur-Grover:
> > Also - New Arch Gallery is now in main - you can merge changes for it into the main branch instead of the NewArch branch.
> 
> Oh I see! So does main branch changes propogate to newarch or we only fix the main?

NewArch is not hosted within main - so the NewArch branch no longer needs to be updated.

PR #591, opened 2025-07-01
[React Native New Architecture]: Incorrect keyboard focus order is observed on invoking the 'Expand menu' button.

HariniMalothu17:
## Description
[React Native New Architecture]: Incorrect keyboard focus order is observed on invoking the 'Expand menu' button.
### Why
On invoking the 'Experimental menu' button, keyboard focus is getting lost. User has to Tab nearly 15-17 times to move the focus to the first menu item.

Expected Result:
Keyboard focus should not get lost on invoking 'Experimental menu' button. It should lands on the first menu item(Home) once after invoking the 'Experimental menu' button.

### What

## Screenshots

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/591)

### DIFF
--- a/NewArch/src/App.tsx
+++ b/NewArch/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import {
   View,
   StyleSheet,
@@ -6,13 +6,14 @@ import {
   Text,
   useColorScheme,
   ScrollView,
+  KeyboardEvent as RNKeyboardEvent,
 } from 'react-native';
-import {NavigationContainer} from './Navigation';
+import { NavigationContainer } from './Navigation';
 import {
   createDrawerNavigator,
   getDrawerStatusFromState,
 } from './Navigation';
-import RNGalleryList, {RNGalleryCategories} from './RNGalleryList';
+import RNGalleryList, { RNGalleryCategories } from './RNGalleryList';
 import LightTheme from './themes/LightTheme';
 import DarkTheme from './themes/DarkTheme';
 import {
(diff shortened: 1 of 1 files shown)


NewArch/src/App.tsx
```diff
@@ -276,28 +281,60 @@ const DrawerListView = (props) => {
...
+      setTimeout(() => {
+        homeRef.current?.focus();
+      }, 300);
+    }
+  }, [isDrawerOpen]);
+
+  // Keyboard navigation looping handlers
+  const onHamburgerKeyDown = (e: RNKeyboardEvent) => {
+    if (e.nativeEvent.key === 'Tab' && e.nativeEvent.shiftKey) {
+      e.preventDefault();
```
acoates-ms, line 308: preventDefault does not prevent the native behavior (The native behavior happenson a different thread and is not blocked by the JS eventing) - You need to set the KeyboardEvents property to disable the native bahavior. https://microsoft.github.io/react-native-windows/docs/ikeyboardprops-api#keydownevents

NewArch/src/App.tsx
```diff
@@ -276,28 +281,60 @@ const DrawerListView = (props) => {
...
-  }
+  // Refs for hamburger, home and settings buttons
+  const hamburgerRef = useRef<Pressable>(null);
+  const homeRef = useRef<Pressable>(null);
+  const settingsRef = useRef<Pressable>(null);
+
+  // When drawer opens, focus the Home menu item
+  useEffect(() => {
+    if (isDrawerOpen && homeRef.current) {
+      setTimeout(() => {
```
acoates-ms, line 299: Requiring a setTimeout is an indication of a synchronization issue.  And the 300ms is arbitrary. 
If we have a ref to home, I think it should already be mounted, and we should be able to call focus on it?

NewArch/src/App.tsx
HariniMalothu17: Resolved

vineethkuttan:
Tried your changes in locally. Working fine for me.

PR #597, opened 2025-07-03
[React Native New Architecture] Fix: The drawer has size constraints

### LINKED ISSUE #595: [React Native New Architecture] The drawer has size constraints that are causing tab navigation to get clipped.
Repro-Steps:
Launch the React Native Gallery- Experimental app.
Navigate to the 'Expand menu' button and invoke it.
Navigate to any of the buttons from the left navigation bar and invoke it.
Observe the keyboard focus and behavior.
The Problem:
When users navigate through the drawer using tab key, focused elements get visually truncated/clipped at the right edge, making them partially invisible and difficult to read.

https://github.com/user-attachments/assets/0512046c-8570-434a-abb8-f2c106555e80

anupriya13:
Resolved #595 
image
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/597)

### DIFF
--- a/NewArch/src/Navigation.tsx
+++ b/NewArch/src/Navigation.tsx
@@ -228,12 +228,9 @@ const DrawerNavigator = ({drawerContent, defaultStatus, children} : DrawerNaviga
       style={{
         flex: 1,
         maxWidth: DEFAULT_DRAWER_WIDTH,
-        paddingLeft: 16, // Add padding for spacing
+        paddingLeft: 16,
+        paddingRight: 6,
       }}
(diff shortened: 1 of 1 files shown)


anupriya13:
> Hi! Sorry just looking at this now. Does this change still work when the text size increases. I know we've been have some other accessibility issues with the drawer when text size gets larger. Can you check if the focus ring can still fully be seen when the text size gets bigger? 

Yes that's verified already

Issue #590, opened 2025-07-01
Incorrect keyboard focus order is observed on invoking any of the buttons from 'Expand menu' left navigation bar.


HariniMalothu17:
Build Details:
OS Version: Dev (OS Build 27881.1000)
App: React Native Gallery Experimental
AT: Keyboard

 

Repro-Steps:
Launch the React Native Gallery- Experimental app.
Navigate to the 'Expand menu' button and invoke it.
Navigate to any of the buttons from the left navigation bar and invoke it.
Observe the keyboard focus and behavior.
Actual Result:
On invoking any of the buttons from the left navigation bar, keyboard focus is getting lost. Again, on hitting Tab focus is moving to the Experimental menu bar and then to 'Experimental Menu' button. Then again on hitting Tab focus is moving to the first interactive element inside the related page.

Expected Result:
Keyboard focus should not get lost on invoking any buttons from the left navigation bar. It should directly lands on the first interactive element inside the page.

User impact:

Users with motor impairment who rely on keyboard will face difficulties and might get confused if keyboard focus got lost on invoking any of the button from the left pane.

MAS Reference: [MAS 2.4.3 – Focus Order](https://aka.ms/MAS2.4.3)
 
Trap ID: [4.1 - Unnecessary Step](https://aka.ms/UnnecessaryStep)
 
“Have feedback on bugs logged, please tag bug as “A11yRCA” and add your feedback in the comment section.”

PR #625, opened 2025-07-16
[React Native New Architecture]Incorrect keyboard focus order is observed on invoking any of the buttons from 'Expand menu' left navigation bar.

HariniMalothu17:
## Description
On invoking any of the buttons from the left navigation bar, keyboard focus is getting lost. Again, on hitting Tab focus is moving to the Experimental menu bar and then to 'Experimental Menu' button. Then again on hitting Tab focus is moving to the first interactive element inside the related page.
### Why
Keyboard focus should not get lost on invoking any buttons from the left navigation bar. It should directly lands on the first interactive element inside the page.

Resolves [#590 ]

Created usePageFocusManagement hook (usePageFocusManagement.ts)
Uses useLayoutEffect for immediate focus after render
Focuses first element on page load and drawer navigation
Updated All Example Pages (20+ files)
Files: ButtonExamplePage.tsx, TextExamplePage.tsx, SwitchExamplePage.tsx, etc.
Added const firstTileRef = usePageFocusManagement(navigation)
Applied ref={firstTileRef} to first interactive element
Result: Focus goes to first button/control instead of hamburger menu

## Screenshots
Before:

https://github.com/user-attachments/assets/9d62a55f-1047-434e-b2dd-48b4e630119e

After:

https://github.com/user-attachments/assets/69388ceb-4026-44f2-bbad-625679ccc849

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/625)

### DIFF
--- a/NewArch/src/ComponentListPage.tsx
+++ b/NewArch/src/ComponentListPage.tsx
@@ -136,6 +136,7 @@ const ComponentListPage = ({route, navigation}: ComponentListPageProps) => {
                 All samples
               </Text>
               <GroupedListOfAllComponents
+                route={route}
                 navigation={navigation}
                 firstTileRef={firstTileRef}
               />
--- a/NewArch/src/components/ControlItem.tsx
+++ b/NewArch/src/components/ControlItem.tsx
@@ -8,6 +8,7 @@ import {
   Image,
 } from 'react-native';
 import React from 'react';
+import type {IRNGalleryExample} from '../RNGalleryList';
 //import {useTheme} from '@react-navigation/native';
 
 const createStyles = (colors: any, isHovered: boolean, isPressing: boolean) =>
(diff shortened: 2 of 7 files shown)


vineethkuttan:
@HariniMalothu17  could you please confirm whether Narrator is correctly narrating the focused component?

NewArch/src/ComponentListPage.tsx
```diff
@@ -120,8 +136,8 @@ const ComponentListPage = ({route, navigation}: ComponentListPageProps) => {
                 All samples
               </Text>
               <GroupedListOfAllComponents
-                route={route}
```
vineethkuttan, line 123: Why route removed?

NewArch/src/components/ControlItem.tsx
```diff
@@ -66,10 +65,11 @@ const createStyles = (colors: any, isHovered: boolean, isPressing: boolean) =>
   });
 
 type HomeComponentTileProps = {
-  item: IRNGalleryExample;
```
vineethkuttan, line 69: Why it is removed? unused?

NewArch/src/examples/VirtualizedListExamplePage.tsx
```diff
@@ -301,7 +303,12 @@ export const VirtualizedListExamplePage: React.FunctionComponent<{}> = () => {
       ]}>
       <Example title="A simple VirtualizedList." code={example1jsx}>
         <ScrollView horizontal={true}>
-          <View style={styles.container}>
+          <View
+            ref={firstVirtualizedListRef}
+            style={styles.container}
+            accessible={true}
+            accessibilityRole="list"
```
vineethkuttan, line 310: accessible and accessibilityRole prop is not necessary, I think.
We are setting list role to the child, so not needed.

HariniMalothu17:
> @HariniMalothu17 could you please confirm whether Narrator is correctly narrating the focused component?

Verified.It's working as expected

PR #738, opened 2025-10-15
Fixing issue where in half screen mode button content is obscured

protikbiswas100:
## Description
In half screen mode button content is obscured
### Why
In half screen mode button content is obscured, all buttons padding, height, width, margin are hardcoded causing the issue when the app window is changed, this change gets the window size before setting height, width, margin etc.

Resolves [https://github.com/microsoft/react-native-gallery/issues/726]

### What
In half screen mode button content is obscured, all buttons padding, height, width, margin are hardcoded causing the issue when the app window is changed, this change gets the window size before setting height, width, margin etc.

## Screenshots

Before
obs-before

After
buttion-after

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-gallery/pull/738)

### DIFF
--- a/NewArch/jest.config.js
+++ b/NewArch/jest.config.js
@@ -1,3 +1,10 @@
 module.exports = {
   preset: 'react-native',
+  transformIgnorePatterns: [
+    'node_modules/(?!(react-native|@react-native|react-native-windows)/)',
+  ],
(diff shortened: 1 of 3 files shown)


Nitin-100 (approved):
LGTM

Nitin-100 (approved):

PR #929, opened 2026-09-13
 Fix app title text scaling on Windows

anuagragith:
## Summary
Fix the React Native Gallery window title so it responds to the Windows Accessibility > Text size setting.

- Render the visible title using React Native with font scaling enabled.
- Preserve `AppWindow.Title` for accessibility, task switching, and window identity.
- Switch the native caption between standard and tall heights as text size changes.
- Keep the React title and native caption synchronized during live text-size changes.
- Preserve native caption buttons, window dragging, system colors, and inactive-window colors.
- Add regression tests for the title content and live title-bar height changes.

## Root cause
The Windows App SDK native caption text does not respond to the system accessibility text-size setting and does not expose an API for changing its font size.

Additionally, RNW 0.84 updates Fabric text layout when text size changes but does not publish the updated `fontScale` through `Dimensions`. The implementation therefore uses the title’s actual rendered height to keep the React content boundary synchronized with the native caption.

## Validation
- Verified live transitions at 100%, 200%, 225%, and back to 100% without restarting.
- Title height changed as expected: 24 → 48 → 54 → 24 pixels.
- Caption-to-content spacing remained consistently 8 pixels with no overlap or excess gap.
- Native minimize, maximize, and close buttons remained available.
- Window title-bar dragging remained functional.
- Full Jest suite passed: 20 tests.
- Full lint completed with 0 errors; existing warnings only.
- Debug x64 Windows build succeeded.
- Independent regression review found no remaining actionable issues.

## Scope
This fixes ADO 63971606 for the main Gallery window.

Note: This is a pre-existing Gallery accessibility issue and was not introduced by React Native Windows 0.85.


anuagragith:
Closing per discussion with Andrew Coates. The standalone non-RNW AppWindow repro confirms that this is a pre-existing external platform issue in the native Windows caption-rendering path, not an RNW 0.85 regression. ADO 63971606 has been marked External and non-blocking for RNW 0.85, and Gallery will remain unchanged. RNW real-time text-scale handling is tracked separately in microsoft/react-native-windows#10253.

